PR #4, opened 2014-08-04
Build oozie with specified hadoop version

watermen:


ajayyadava:
@watermen We have started to accept pull requests but we still use JIRA for issue tracking. Would you mind creating a JIRA and updating your pull request?

PR #10, opened 2015-12-16
[FALCON-1669] add printStackTrace before throw FalconCLIException

baishuo:
I think we should print the stacktrace to let user know the root cause of why  Falcon Client object starts fail.

### DIFF
--- a/client/src/main/java/org/apache/falcon/client/FalconClient.java
+++ b/client/src/main/java/org/apache/falcon/client/FalconClient.java
@@ -150,6 +150,7 @@ public class FalconClient extends AbstractFalconClient {
             client.resource(UriBuilder.fromUri(baseUrl).build());
             authenticationToken = getToken(baseUrl);
         } catch (Exception e) {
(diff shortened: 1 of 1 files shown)


client/src/main/java/org/apache/falcon/client/FalconClient.java
```diff
@@ -150,6 +150,7 @@ public FalconClient(String falconUrl, Properties properties) throws FalconCLIExc
             client.resource(UriBuilder.fromUri(baseUrl).build());
             authenticationToken = getToken(baseUrl);
         } catch (Exception e) {
+        	e.printStackTrace();
```
pavankumar526: It was changed like this throw new FalconCLIException("Unable to initialize Falcon Client object. Cause : " + e.getMessage(), e); 

@baishuo  Can you please close this pull request if it is no longer required.

ajayyadava:
@baishuo Do you mind closing this pull request?

PR #14, opened 2016-01-27
FALCON-1609: Add entity commands to falcon spring-shell CLI

prongs:


ajayyadava:
@prongs  I suspect you need to mention JIRA id as FALCON-1609 (note the capital case characters) so that the comments are posted on the corresponding JIRA.

cli/src/main/java/org/apache/falcon/cli/commands/FalconEntityCommands.java
```diff
@@ -18,9 +18,303 @@
...
+import static org.apache.falcon.cli.FalconEntityCLI.UPDATE_OPT;
+import static org.apache.falcon.cli.FalconEntityCLI.UPDATE_OPT_DESCRIPTION;
+import static org.apache.falcon.cli.FalconEntityCLI.VALIDATE_OPT;
+import static org.apache.falcon.cli.FalconEntityCLI.VALIDATE_OPT_DESCRIPTION;
+import static org.apache.falcon.cli.FalconEntityCLI.validateEntityFields;
+
 /**
  * Entity Commands.
  */
+// SUSPEND CHECKSTYLE CHECK ParameterNumberCheck
```
ajayyadava: Instead of suspending check for entire class, how about moving it to specific methods?

client/src/main/java/org/apache/falcon/cli/FalconCLI.java
```diff
@@ -48,8 +48,10 @@
 
     public static final String ENV_FALCON_DEBUG = "FALCON_DEBUG";
     public static final String DEBUG_OPTION = "debug";
+    public static final String DEBUG_OPTION_DESCRIPTION = "Use debug mode to see debugging statements on stdout";
```
ajayyadava: Most of these are relevant to only shell, so does it make sense to move to that module?

client/src/main/java/org/apache/falcon/client/FalconCLIException.java
```diff
@@ -27,7 +27,7 @@
 /**
  * Exception thrown by FalconClient.
  */
-public class FalconCLIException extends Exception {
+public class FalconCLIException extends RuntimeException{
```
ajayyadava: It will be worthwhile to add a 1 line comment on why it was done for future reference.

client/src/main/java/org/apache/falcon/cli/FalconCLI.java
ajayyadava: Ignore my previous comment. It is better to keep the option and its description together.

ajayyadava:
In Falcon we maintain all change log in CHANGES.txt. Since we don't make edits to the commits sent by the users you will also need to update CHANGES.txt (top level)

ajayyadava:
I haven't reviewed all the command options in detail (mandatory vs. optional parameters) so I hope they are consistent with the old ones. Apart from that and minor nits, the change looks good to me.

client/src/main/java/org/apache/falcon/cli/FalconCLI.java
prongs: Yes, can be moved in the switch.

CHANGES.txt
```diff
@@ -51,6 +51,10 @@ Trunk
 
     FALCON-1779 Fix rat-check failure in trunk (Ajay Yadava)
 
+  NEW FEATURES
```
ajayyadava: This needs to be removed as we have stopped generating changes manually.

cli/src/main/java/org/apache/falcon/cli/commands/FalconEntityCommands.java
```diff
@@ -18,9 +18,304 @@
...
  * Entity Commands.
  */
+@Component
 public class FalconEntityCommands extends BaseFalconCommands {
+    public static final String ENTITY_PREFIX = "entity";
+    public static final String ENTITY_COMMAND_PREFIX = ENTITY_PREFIX + " ";
+
+    @CliCommand(value = ENTITY_COMMAND_PREFIX + SLA_MISS_ALERT_OPT, help = SLA_MISS_ALERT_OPT_DESCRIPTION)
+    public String slaAlert(
+            @CliOption(key = {TYPE_OPT}, mandatory = true, help = TYPE_OPT_DESCRIPTION) final EntityType entityType,
```
ajayyadava: currently key can only be Feed, so we need to validate that.

cli/src/main/java/org/apache/falcon/cli/commands/FalconEntityCommands.java
```diff
@@ -18,9 +18,304 @@
...
+            @CliOption(key = {ENTITY_NAME_OPT}, mandatory = true, help = ENTITY_NAME_OPT_DESCRIPTION)
+            final String entityName,
```
ajayyadava: entityName is optional.

cli/src/main/java/org/apache/falcon/cli/commands/FalconEntityCommands.java
```diff
@@ -18,9 +18,304 @@
...
+            @CliOption(key = {COLO_OPT}, mandatory = true, help = COLO_OPT_DESCRIPTION) final String colo,
+            @CliOption(key = {START_OPT}, mandatory = false, help = START_OPT_DESCRIPTION) final String start,
```
ajayyadava: start is mandatory, end is optional.

cli/src/main/java/org/apache/falcon/cli/commands/FalconEntityCommands.java
```diff
@@ -18,9 +18,304 @@
...
+            @CliOption(key = {TYPE_OPT}, mandatory = true, help = TYPE_OPT_DESCRIPTION) final EntityType entityType,
+            @CliOption(key = {FILE_PATH_OPT}, mandatory = true, help = FILE_PATH_OPT_DESCRIPTION) final File filePath
+    ) {
+
+        return getFalconClient().submit(entityType.name().toLowerCase(), filePath.getPath(), getDoAs()).getMessage();
+    }
+
+    @CliCommand(value = ENTITY_COMMAND_PREFIX + LOOKUP_OPT, help = LOOKUP_OPT_DESCRIPTION)
+    public String lookup(
+            @CliOption(key = {TYPE_OPT}, mandatory = true, help = TYPE_OPT_DESCRIPTION) final EntityType entityType,
```
ajayyadava: I think in general, there is no validation happening on FalconClient, so we should do validation in shell/cli. e.g. entityType here can only be feed. What do you think?

cli/src/main/java/org/apache/falcon/cli/commands/FalconEntityCommands.java
```diff
@@ -18,9 +18,304 @@
...
+    }
+
+    @CliCommand(value = ENTITY_COMMAND_PREFIX + UPDATE_OPT, help = UPDATE_OPT_DESCRIPTION)
+    public String update(
+            @CliOption(key = {TYPE_OPT}, mandatory = true, help = TYPE_OPT_DESCRIPTION) final EntityType entityType,
+            @CliOption(key = {ENTITY_NAME_OPT}, mandatory = true, help = ENTITY_NAME_OPT_DESCRIPTION)
+            final String entityName,
+            @CliOption(key = {FILE_PATH_OPT}, mandatory = true, help = FILE_PATH_OPT_DESCRIPTION)
+            final File filePath,
+            @CliOption(key = {SKIPDRYRUN_OPT}, mandatory = false, help = SKIPDRYRUN_OPT_DESCRIPTION,
```
ajayyadava: mandatory field.

cli/src/main/java/org/apache/falcon/cli/commands/FalconEntityCommands.java
```diff
@@ -18,9 +18,304 @@
...
+                .getMessage();
+    }
+
+    @CliCommand(value = ENTITY_COMMAND_PREFIX + SUBMIT_AND_SCHEDULE_OPT, help = SUBMIT_AND_SCHEDULE_OPT_DESCRIPTION)
+    public String submitAndSchedule(
+            @CliOption(key = {TYPE_OPT}, mandatory = true, help = TYPE_OPT_DESCRIPTION) final EntityType entityType,
+            @CliOption(key = {FILE_PATH_OPT}, mandatory = true, help = FILE_PATH_OPT_DESCRIPTION) final File filePath,
+            @CliOption(key = {SKIPDRYRUN_OPT}, mandatory = false, help = SKIPDRYRUN_OPT_DESCRIPTION,
+                    unspecifiedDefaultValue = "false", specifiedDefaultValue = "true") boolean skipDryRun,
+            @CliOption(key = {PROPS_OPT}, mandatory = true, help = PROPS_OPT_DESCRIPTION) final String properties
```
ajayyadava: properties are optional.

cli/src/main/java/org/apache/falcon/cli/commands/FalconEntityCommands.java
```diff
@@ -18,9 +18,304 @@
...
+        return getFalconClient()
+                .validate(entityType.name().toLowerCase(), filePath.getPath(), skipDryRun, getDoAs())
+                .getMessage();
+    }
+
+    @CliCommand(value = ENTITY_COMMAND_PREFIX + SCHEDULE_OPT, help = SCHEDULE_OPT_DESCRIPTION)
+    public String schedule(
+            @CliOption(key = {TYPE_OPT}, mandatory = true, help = TYPE_OPT_DESCRIPTION) final EntityType entityType,
+            @CliOption(key = {ENTITY_NAME_OPT}, mandatory = true, help = ENTITY_NAME_OPT_DESCRIPTION) String entityName,
+            @CliOption(key = {COLO_OPT}, mandatory = true, help = COLO_OPT_DESCRIPTION) final String colo,
```
ajayyadava: I think colo is optional for most of the commands and if null a default value of * is used.

cli/src/main/java/org/apache/falcon/cli/commands/FalconEntityCommands.java
```diff
@@ -18,9 +18,304 @@
...
+    ) {
+
+        return getFalconClient().schedule(entityType, entityName, colo, skipDryRun, getDoAs(), properties).getMessage();
+    }
+
+    @CliCommand(value = ENTITY_COMMAND_PREFIX + SUSPEND_OPT, help = SUSPEND_OPT_DESCRIPTION)
+    public String suspend(
+            @CliOption(key = {TYPE_OPT}, mandatory = true, help = TYPE_OPT_DESCRIPTION) final EntityType entityType,
+            @CliOption(key = {ENTITY_NAME_OPT}, mandatory = true, help = ENTITY_NAME_OPT_DESCRIPTION) String entityName,
+            @CliOption(key = {COLO_OPT}, mandatory = true, help = COLO_OPT_DESCRIPTION) final String colo
```
ajayyadava: optional with default value of *

cli/src/main/java/org/apache/falcon/cli/commands/FalconEntityCommands.java
```diff
@@ -18,9 +18,304 @@
...
+    @CliCommand(value = ENTITY_COMMAND_PREFIX + SUMMARY_OPT, help = SUMMARY_OPT_DESCRIPTION)
+    public String summary(
+            @CliOption(key = {TYPE_OPT}, mandatory = true, help = TYPE_OPT_DESCRIPTION) final EntityType entityType,
+            @CliOption(key = {CLUSTER_OPT}, mandatory = true, help = CLUSTER_OPT_DESCRIPTION) final String cluster,
+            @CliOption(key = {START_OPT}, mandatory = true, help = START_OPT_DESCRIPTION) final String start,
+            @CliOption(key = {END_OPT}, mandatory = true, help = END_OPT_DESCRIPTION) final String end,
+            @CliOption(key = {FIELDS_OPT}, mandatory = true, help = FIELDS_OPT_DESCRIPTION) final String fields,
+            @CliOption(key = {ORDER_BY_OPT}, mandatory = true, help = ORDER_BY_OPT_DESCRIPTION) final String orderBy,
+            @CliOption(key = {SORT_ORDER_OPT}, mandatory = true,
+                    help = SORT_ORDER_OPT_DESCRIPTION) final String sortOrder,
```
ajayyadava: all the remaining parameters in this method are also optional I believe.

client/src/main/java/org/apache/falcon/client/FalconClient.java
```diff
@@ -83,7 +83,7 @@
  */
 public class FalconClient extends AbstractFalconClient {
 
-    public static final AtomicReference<PrintStream> OUT = new AtomicReference<>(System.out);
+    public static final AtomicReference<PrintStream> OUT = new AtomicReference<PrintStream>(System.out);
```
ajayyadava: Is it required?

cli/src/main/java/org/apache/falcon/cli/commands/FalconEntityCommands.java
prongs: Is there no validation in `FalconEntityCLI` for that? 
```
if (optionsList.contains(SLA_MISS_ALERT_OPT)) {
            validateNotEmpty(entityType, TYPE_OPT);
            validateNotEmpty(start, START_OPT);
            parseDateString(start);
            parseDateString(end);
            SchedulableEntityInstanceResult response = client.getFeedSlaMissPendingAlerts(entityType,
                    entityName, start, end, colo);
            result = ResponseHelper.getString(response);
```

cli/src/main/java/org/apache/falcon/cli/commands/FalconEntityCommands.java
prongs: I believe such a validation should be done in server. That way the validation will be centralized. Otherwise all clients need to add validation code themselves. Given the clients can be in java/python/node/ruby/..., each needs to implement validations.

client/src/main/java/org/apache/falcon/client/FalconClient.java
prongs: Reverted.

bvellanki:
@prongs : The build failed again. Before you submit the patch, can you please run the following command to ensure build works 

"mvn clean verify assembly:assembly -Ptest-patch"

prongs:
Build failed because of checkstyle issues. You can continue to review otherwise. Next commit I add will just be fixing checkstyle problems.

prongs:
@bvellanki Can you review?

cli/src/main/java/org/apache/falcon/cli/FalconEntityCLI.java
```diff
@@ -259,17 +288,21 @@ public void entityCommand(CommandLine commandLine, FalconClient client) throws F
...
+            validateColo(optionsList);
+            result = client.submitAndSchedule(entityType, filePath, skipDryRun, doAsUser, userProps).getMessage();
         }  else if (optionsList.contains(FalconCLIConstants.UPDATE_CLUSTER_DEPENDENTS_OPT)) {
             validateNotEmpty(cluster, FalconCLIConstants.CLUSTER_OPT);
             result = client.updateClusterDependents(cluster, skipDryRun, doAsUser).getMessage();
+        } else if (optionsList.contains(VALIDATE_OPT)) {
+            validateNotEmpty(filePath, FILE_PATH_OPT);
+            validateColo(optionsList);
+            result = client.validate(entityType, filePath, skipDryRun, doAsUser).getMessage();
         } else if (optionsList.contains(FalconCLIConstants.SUBMIT_AND_SCHEDULE_OPT)) {
```
bvellanki: This is duplicate code, refer to lines 291-294 above

cli/src/main/java/org/apache/falcon/cli/FalconEntityCLI.java
```diff
@@ -329,13 +362,13 @@ public void entityCommand(CommandLine commandLine, FalconClient client) throws F
         OUT.get().println(result);
     }
 
-    private void validateColo(Set<String> optionsList) throws FalconCLIException {
+    private void validateColo(Set<String> optionsList) {
```
bvellanki: This method throws FalconCLIException. Why is it removed from method signature? Same applies for most methods below.

prongs: That's a RuntimeException. Checkstyle was failing because of that.

cli/src/main/java/org/apache/falcon/cli/FalconCLI.java
```diff
@@ -48,7 +48,45 @@
     public static final AtomicReference<PrintStream> ERR = new AtomicReference<PrintStream>(System.err);
     public static final AtomicReference<PrintStream> OUT = new AtomicReference<PrintStream>(System.out);
 
-    private static final String FALCON_URL = "FALCON_URL";
+    public static final String ENV_FALCON_DEBUG = "FALCON_DEBUG";
+    public static final String DEBUG_OPTION = "debug";
+    public static final String URL_OPTION = "url";
+    public static final String FALCON_URL = "FALCON_URL";
+
```
bvellanki: Most, if not all of these are moved to FalconCLIConstants.java by FALCON-1895

bvellanki:
@prongs : I left some comments.  Looks like FALCON-1895 has changed CLI significantly and your pull request  might have to change significantly as well.

cli/src/main/java/org/apache/falcon/cli/FalconEntityCLI.java
prongs: Not true. not empty and colo validations are being done in validation of other options too. I agree that the middle two lines are same, but abstracting just two lines out in a method would be over-engineering.

prongs:
No. Currently, I'm seeing no conflicts with the base branch. FALCON-1895 has been committed since April. There are no conflicts.

cli/src/main/java/org/apache/falcon/cli/FalconEntityCLI.java
bvellanki: Checkstyle was probably failing due to a different issue. The existing code has "throws FalconCLIException" and it does not throw checkstyle errors.

cli/src/main/java/org/apache/falcon/cli/FalconCLI.java
bvellanki: SUBMIT_AND_SCHEDULE_OPT is handled twice in the merged code. Once in lines 291 to 294, again in line 302-305.

cli/src/main/java/org/apache/falcon/cli/FalconEntityCLI.java
prongs: I was clearly getting `Redundant throws` on both local and jenkins. Some of the builds have failed with that. I could prove it to you but the job logs would have been lost.

bvellanki:
@prongs  It is true there are no merge conflicts, but the patch provided currently is not in alignment with the existing code after FALCON-1895 refactoring.

prongs:
@bvellanki: 


The refactoring done in FALCON-1895 might not be long lived. I have a bigger picture in mind. There will be a client module which will empower both the existing `cli` and the new `cli` in development. The existing `cli` will still be called `cli`, while I intend to rename the new cli to `shell`. (Notice that the code in this pull request is still living in the `cli` module.) With that, the `client` module will empower two modules. The constants class that has been extracted right now contains only the string constants for option names, and option descriptions are not refactored. That seems incomplete to me on the one hand, while on the other hand, I've seen instances where multiple CLI subclasses are using the same option names but different option descriptions. So I'm reluctant to empower that class right now. So, I'll use that to the extent that duplication of constants is removed, but no more. 

Secondly, I talked about the bigger picture. I have completely developed the new cli in #14, #52, #53, #54 and #55. They have been there for a little over 3 months now without anyone to review. My intention is to get those merged, then separate the modules out, and then do a final refactoring like FALCON-1895. What you are suggesting will need me to do a major refactoring in each of those pull requests. That will be a slow process for me since I don't work full-time on Falcon. Whereas making minor changes would take minimal time, and that I can afford. And even with refactoring each of those, a final refactoring might still be required. 

So, in favor of faster development of the shell, I'm proposing that we first move forward with what's already there (#14, #52, #53, #54, #55) and then clean up later. After the clean-up, I intend to unify the `cli` and `shell` through a shell script, where typing `falcon ...` will go through the CLI path, but just typing `falcon` will open the `shell`.

ajayyadava:
Sorry for the churn on this and thank you for taking this up @prongs. This is a very valuable addition to the project.

I agree with @bvellanki 's observations but I am satisfied with @prongs approach. Things may seem bit unorganized but for sake of smaller commits and moving forward I think this is the correct approach. Also, it shouldn't impact any of the existing functionality so is a low risk change. This has been pending for quite some time and couple of iterations have already been done without any blocker issues, so I will commit this after 24 hours unless someone shouts out :-)

PR #15, opened 2016-01-28
server side check added

PraveenAdlakha:
n/falcon entity -type feed -submitAndSchedule -file ../cluster.xml 
ERROR: Forbidden;submitAndSchedule on server is not supported.Please run your operation on Prism.

$ bin/falcon entity -type feed -update -name abc -file ../cluster.xml 
ERROR: Forbidden;update on server is not supported.Please run your operation on Prism.

$ bin/falcon entity -type feed -delete -name abc 
ERROR: Forbidden;delete on server is not supported.Please run your operation on Prism.

$ bin/falcon entity -type feed -submit -file ../cluster.xml 
ERROR: Forbidden;submit on server is not supported.Please run your operation on Prism.

### DIFF
--- a/client/src/main/java/org/apache/falcon/client/FalconCLIException.java
+++ b/client/src/main/java/org/apache/falcon/client/FalconCLIException.java
@@ -47,7 +47,6 @@ public class FalconCLIException extends Exception {
         ClientResponse.Status status = clientResponse.getClientResponseStatus();
         String statusValue = status.toString();
         String message = "";
-        if (status == ClientResponse.Status.BAD_REQUEST) {
             clientResponse.bufferEntity();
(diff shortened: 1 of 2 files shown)


client/src/main/java/org/apache/falcon/client/FalconCLIException.java
```diff
@@ -47,7 +47,6 @@ public static FalconCLIException fromReponse(ClientResponse clientResponse) {
         ClientResponse.Status status = clientResponse.getClientResponseStatus();
         String statusValue = status.toString();
         String message = "";
-        if (status == ClientResponse.Status.BAD_REQUEST) {
```
sepreece: You need to adjust the indentation on the lines that used to be in the block.

PR #16, opened 2016-01-28
FALCON-1774 Server side check added

PraveenAdlakha:
Testing :

$ bin/falcon entity -type feed -submitAndSchedule -file ../cluster.xml 
ERROR: Forbidden;submitAndSchedule on server is not supported.Please run your operation on Prism.

$ bin/falcon entity -type feed -update -name abc -file ../cluster.xml 
ERROR: Forbidden;update on server is not supported.Please run your operation on Prism.

$ bin/falcon entity -type feed -delete -name abc 
ERROR: Forbidden;delete on server is not supported.Please run your operation on Prism.

$ bin/falcon entity -type feed -submit -file ../cluster.xml 
ERROR: Forbidden;submit on server is not supported.Please run your operation on Prism.

### DIFF
--- a/webapp/src/main/java/org/apache/falcon/resource/SchedulableEntityManager.java
+++ b/webapp/src/main/java/org/apache/falcon/resource/SchedulableEntityManager.java
@@ -27,8 +27,10 @@ import javax.ws.rs.Path;
 import javax.ws.rs.PathParam;
 import javax.ws.rs.Produces;
 import javax.ws.rs.QueryParam;
+import javax.ws.rs.DELETE;
 import javax.ws.rs.core.Context;
 import javax.ws.rs.core.MediaType;
+import javax.ws.rs.core.Response;
 
 import org.apache.commons.lang.StringUtils;
 import org.apache.falcon.FalconWebException;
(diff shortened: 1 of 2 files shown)


webapp/src/main/java/org/apache/falcon/resource/SchedulableEntityManager.java
```diff
@@ -57,6 +59,91 @@ public APIResult getStatus(@Dimension("entityType") @PathParam("type") String ty
...
+    @DELETE
+    @Path("delete/{type}/{entity}")
+    @Produces({MediaType.TEXT_XML, MediaType.TEXT_PLAIN, MediaType.APPLICATION_JSON})
+    @Monitored(event = "delete")
+    @Override
+    public APIResult delete(
+            @Context HttpServletRequest request, @Dimension("entityType") @PathParam("type") final String type,
+            @Dimension("entityName") @PathParam("entity") final String entity,
+            @Dimension("colo") @QueryParam("colo") String ignore) {
+        throw FalconWebException.newAPIException("delete on server is not" +
```
pallavi-rao: Shouldn't we check if it is embedded mode or not?

PraveenAdlakha: Can you please help me understand why we want to check this?
As per my understanding these calls are not supported on falcon server and whenever this method gets called we should through an error.

pallavi-rao: Falcon can run in 2 modes: 
1. Distributed. In this case, Falcon prism is what user should use to perform CRUD operations on entities and direct CRUD operations against Falcon Server should not be allowed.
2. Embedded. In this case, there is no Prism server at all, there is just one instance of Server and all CRUD operations should be allowed against server in this mode.

PraveenAdlakha: Hi ,

I have confirmed its working fine in embedded mode it checks intoSchedulableEntityManageProxy class not in SchedulableEntityManager :

$  bin/falcon entity -type process -name DWH-CI-daily-supply5 -delete
falcon/default/DWH-CI-daily-supply5(process) removed successfully (KILLED in WF_ENGINE)

pallavi-rao:
+1. LGTM. The build needs to pass though before merge.

ajayyadava:
I have retriggered the build (https://builds.apache.org/job/Apache-Falcon-Pull-Request-Build/16/)

PR #17, opened 2016-01-29
FALCON-1750 Null Pointer Exception while listening to Workflow Notifi…

pallavi-rao:
…cations

### DIFF
--- a/common/src/main/java/org/apache/falcon/workflow/WorkflowJobEndNotificationService.java
+++ b/common/src/main/java/org/apache/falcon/workflow/WorkflowJobEndNotificationService.java
@@ -181,7 +181,7 @@ public class WorkflowJobEndNotificationService implements FalconService {
                                 Integer.toString(instances[0].getRunId()));
                     }
                 } catch (FalconException e) {
-                    // Do Nothing. The workflow may not have been deployed on this cluster.
+                    // Do Nothing. Move on to the next cluster.
                     continue;
                 }
                 contextMap.put(context.getWorkflowId(), wfProps);
--- a/common/src/main/java/org/apache/falcon/workflow/WorkflowJobEndNotificationService.java
+++ b/common/src/main/java/org/apache/falcon/workflow/WorkflowJobEndNotificationService.java
@@ -21,10 +21,9 @@ package org.apache.falcon.workflow;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.falcon.FalconException;
 import org.apache.falcon.aspect.GenericAlert;
+import org.apache.falcon.entity.EntityNotRegisteredException;
 import org.apache.falcon.entity.EntityUtil;
-import org.apache.falcon.entity.store.ConfigurationStore;
 import org.apache.falcon.entity.v0.Entity;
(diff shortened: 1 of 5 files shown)


common/src/main/java/org/apache/falcon/workflow/WorkflowJobEndNotificationService.java
```diff
@@ -152,48 +157,48 @@ public void notifyWait(WorkflowExecutionContext context) {
...
+                            .getJobDetails(cluster, context.getWorkflowId()).getInstances();
+                    if (instances != null && instances.length > 0) {
+                        wfProps = getWFProps(instances[0].getWfParams());
+                        // Required by RetryService. But, is not part of conf.
+                        wfProps.setProperty(WorkflowExecutionArgs.RUN_ID.getName(),
+                                Integer.toString(instances[0].getRunId()));
                     }
-                    contextMap.put(context.getWorkflowId(), wfProps);
+                } catch (FalconException e) {
+                    // Do Nothing. The workflow may not have been deployed on this cluster.
```
sandeepSamudrala: can this state not be caught?for if it was scheduled on this cluster easily? rather than catching it as an exception?.

rerun/src/main/java/org/apache/falcon/rerun/handler/RetryHandler.java
```diff
@@ -75,6 +76,9 @@ public void handleRerun(String clusterName, String entityType, String entityName
                         "All retry attempt failed out of configured: "
                                 + attempts + " attempt for entity instance::");
             }
+        } catch (EntityNotRegisteredException ee) {
```
pavankumar526: I think we should fix this in few other places also. One place i think of is SLAMonitoringHandler for onSuccess  method

pavankumar526: In RetryHandler Failure Case also this exception might occur, later after 0.9 we should fix this cleanly, instead of handling in each and every listener. Even though i am not sure what's the best way to handle this cleanly.

pallavi-rao: Most other listeners other than the ones under rerun seem to just read off the context and not really query the config store for the entity for which notification is generated. So, nothing needs to be fixed immediately.

Agree that we have to have cleaner solution.

common/src/main/java/org/apache/falcon/workflow/WorkflowJobEndNotificationService.java
pallavi-rao: The comment is wrong. Since we get the clusters for a given entity on that colo, this should not happen at all. In fact, the whole try-catch block is superfluous. Will remove.

rerun/src/main/java/org/apache/falcon/rerun/handler/RetryHandler.java
pavankumar526: +1 looks good.

sandeepSamudrala:
LGMT. :+1:

PR #18, opened 2016-01-29
FALCON-1782 Client returns FalconWebException instead of the expected…

PraveenAdlakha:
testing :

$ bin/falcon entity -type process -name process1 -properties falcon.scheduler:native -schedule
ERROR: Bad Request;default/Entity process1 is already scheduled on configured workflow engine.


sandeepSamudrala:
looks good.  :+1:

ajayyadava:
:+1: Will commit shortly.

PR #20, opened 2016-02-01
FALCON-1793 : Feed element action="archive" is submittable via command line tool falcon

deepakbarr:
Submit command output - 

$ bin/falcon entity -submit -type feed -file examples/entity/filesystem/out-feed.xml
ERROR: Bad Request;local/ARCHIVE action type is not supported for retention


pallavi-rao:
We should also update feed-0.1.xsd to keep only "delete" and remove the rest. If that is done an explicit validateFeedRetention may not be required.

pallavi-rao:
Please also update CHANGES.txt

deepakbarr:
I agree @pallavi-rao  !

deepakbarr:
Closing the review request. Will create new.

ajayyadava:
@deepakbarr  you don't need to close the pull request, you can just update the same pull request, the build will also trigger automatically.

PR #33, opened 2016-02-08
FALCON-1821 Update git pull merge script to accept and update JIRA type

pallavi-rao:


falcon_merge_pr.py
```diff
@@ -337,22 +337,45 @@ def resolve_jira_issues(title, merge_branches, comment):
...
-    >>> standardize_jira_ref("%s-1032. Test message with dot after id and ellipsis, ..." % PROJECT_NAME)
-    'FALCON-1032 Test message with dot after id and ellipsis, ...'
-    >>> standardize_jira_ref("%s-6250 %s-6146 %s-5911: Test multiple commit messages." % (PROJECT_NAME, PROJECT_NAME, CAPITALIZED_PROJECT_NAME))
-    'FALCON-6250 FALCON-6146 FALCON-5911 Test multiple commit messages.'
-    >>> standardize_jira_ref("Message without JIRA id")
-    'Message without JIRA id'
-    >>> standardize_jira_ref("[FALCON-1009] id in brackets")
-    'FALCON-1009 id in brackets'
-    """
+    # """
```
ajayyadava: You don't need to append # here.

ajayyadava:
I think this information is already there in JIRA and should be fetched from JIRA. Maintaining same information in two places can cause discrepancies. Let's keep JIRA as source of truth and fetch information from there.

pallavi-rao:
The reason this is added is because of the following reasons:

1. JIRA Type may not directly map to our categorization, for example, backward incompatible changes. Sub-Task can be a "New feature". So, this lets the committer override the default categorization, if required.
2. It is just part of commit message similar to title, contributor (author) that also can be retrieved from JIRA. It is just making it available in one place as git log.

ajayyadava:
In view of FALCON-2063 I believe this change should also not be required.

ajayyadava:
As noticed in my earlier comment, this change is not required any longer as we have started using Apache Yetus. I will be closing this pull request as part of the clean up.

PR #34, opened 2016-02-08
FALCON-1822 Script to generate CHANGES.txt

pallavi-rao:
The script ensures categorization of changes, attribution to contributors and maintains history. It does the following:
1. Given a git tag and release version, it retrieves git log from the tag till HEAD (of the branch you are in).
2. Classifies the change log based on category (uses commit message for that).
3. Appends the new change log on top of old CHANGES.txt.
4. Creates a new CHANGES.txt.new file.
5. During release (or otherwise too), you can now check this new file in, in place of old one. (Will update the confluence page)

Example change log :
`FALCON Change Log
----------------

Release 0.9

BUGS
=======================================
  FALCON-1795 Kill api not killing waiting/ready instances
  sandeep <sandysmdl@gmail.com>
  Commit: 49050c8, github.com/apache/falcon/pull/22`

### DIFF
--- a/falcon_generate_changelist.py
+++ b/falcon_generate_changelist.py
@@ -33,8 +33,8 @@ import re
 import glob
 
 ## Set the below variables before generating the CHANGES
-NEW_RELEASE_VERSION = "0.9"
-PREV_RELEASE_GIT_TAG = "release-0.8-rc0"
+NEW_RELEASE_VERSION = "0.10-SNAPSHOT"
+PREV_RELEASE_GIT_TAG = "release-0.9-rc0"
 CAPITALIZED_PROJECT_NAME = "falcon".upper()
 
 CHANGELIST = "CHANGES.txt"
@@ -99,40 +99,24 @@ for h in hashes:
     category = "BUG"
     if match:
         category = match.group(1).upper()
-        print "Subject : %s, Category : %s " % (subject, category)
         subject = re.sub("\[(\w+)\]", "", subject)
     file_name = TMP_CHANGELIST + "_" + category
     if not (os.path.isfile(file_name)):
         append_to_changelist(file_name, category+"S")
         append_to_changelist(file_name, "=======================================")
-    if "Merge pull" in subject:
-        # Parse old format commit message
-        append_to_changelist(file_name, "  %s" % subject)
-        append_to_changelist(file_name, "  [%s]" % body_lines[0])
-        append_to_changelist(file_name, "")
 
-    elif "maven-release" not in subject:
-        # Parse new format commit message
-        # Get authors from commit message, committer otherwise
-        authors = [committer]
-        if "Author:" in body:
-            authors = [line.split(":")[1].strip() for line in body_lines if "Author:" in line]
+    # Generate GitHub PR URL for easy access if possible
+    github_url = ""
+    if "Closes #" in body:
+        pr_num = [line.split()[1].lstrip("#") for line in body_lines if "Closes #" in line][0]
+        github_url = "github.com/apache/falcon/pull/%s" % pr_num
 
-        # Generate GitHub PR URL for easy access if possible
-        github_url = ""
-        if "Closes #" in body:
-            pr_num = [line.split()[1].lstrip("#") for line in body_lines if "Closes #" in line][0]
-            github_url = "github.com/apache/falcon/pull/%s" % pr_num
-
-        append_to_changelist(file_name, "  %s" % subject)
-        append_to_changelist(file_name, "  %s" % ', '.join(authors))
-        # for author in authors:
-        #     append_to_changelist("  %s" % author)
-        if len(github_url) > 0:
-            append_to_changelist(file_name, "  Commit: %s, %s" % (h, github_url))
-        else:
-            append_to_changelist(file_name, "  Commit: %s" % h)
-        append_to_changelist(file_name, "")
+    append_to_changelist(file_name, "  %s" % subject)
+    if len(github_url) > 0:
+        append_to_changelist(file_name, "  Commit: %s, %s" % (h, github_url))
+    else:
+        append_to_changelist(file_name, "  Commit: %s" % h)
+    append_to_changelist(file_name, "")
 
 append_to_changelist(TMP_CHANGELIST, "FALCON Change Log")
 append_to_changelist(TMP_CHANGELIST, "----------------")
--- a/falcon_generate_changelist.py
+++ b/falcon_generate_changelist.py
@@ -1,0 +1,148 @@
+#!/usr/bin/python
+
+#
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
+# (the "License"); you may not use this file except in compliance with
+# the License.  You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# Creates CHANGES.txt from git history.
+# "Inspired by" https://github.com/apache/FALCON/blob/master/dev/create-release/generate-changelist.py
+#
+# Usage:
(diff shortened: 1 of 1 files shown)


falcon_generate_changelist.py
```diff
@@ -0,0 +1,148 @@
...
+        # Generate GitHub PR URL for easy access if possible
+        github_url = ""
+        if "Closes #" in body:
+            pr_num = [line.split()[1].lstrip("#") for line in body_lines if "Closes #" in line][0]
+            github_url = "github.com/apache/falcon/pull/%s" % pr_num
+
+        append_to_changelist(file_name, "  %s" % subject)
+        append_to_changelist(file_name, "  %s" % ', '.join(authors))
+        # for author in authors:
+        #     append_to_changelist("  %s" % author)
```
sandeepSamudrala: authors required?

pallavi-rao: Yep. We wanted to ensure the right people got credit (or got blamed ;-) )

ajayyadava: Maintaining author in CHANGES.txt is not a good practice, change log should contain only the changes and no credits, we attribute commits to their authors now. Earlier the contributors were not authors of the commits and hence we used to attribute like this.

pallavi-rao: That is true. Now the commit can be easily traced back to the contributor. So, this may not be really necessary. Was just keeping it inline with our previous CHANGES.txt and also since the falcon_merge_pr.py allows for multiple authors, if there is more than one author, thought it will get called out here too.

No personal preference. Can knock if off.

sandeepSamudrala:
:+1: Directionally looks good.

falcon_generate_changelist.py
pallavi-rao: Thanks @sandeepSamudrala . @ajayyadava, do you still want me to remove authors list from changelog?

ajayyadava: There can be multiple reviewers but author of a commit is only one.

falcon_generate_changelist.py
```diff
@@ -0,0 +1,148 @@
...
+    category = "BUG"
+    if match:
+        category = match.group(1).upper()
+        print "Subject : %s, Category : %s " % (subject, category)
+        subject = re.sub("\[(\w+)\]", "", subject)
+    file_name = TMP_CHANGELIST + "_" + category
+    if not (os.path.isfile(file_name)):
+        append_to_changelist(file_name, category+"S")
+        append_to_changelist(file_name, "=======================================")
+    if "Merge pull" in subject:
```
ajayyadava: This part is not applicable to us.

falcon_generate_changelist.py
```diff
@@ -0,0 +1,148 @@
...
+        # Parse old format commit message
+        append_to_changelist(file_name, "  %s" % subject)
+        append_to_changelist(file_name, "  [%s]" % body_lines[0])
+        append_to_changelist(file_name, "")
+
+    elif "maven-release" not in subject:
```
ajayyadava: We don't ever put "maven-release" in commit messages either, so I think we can skip this check.

falcon_generate_changelist.py
pallavi-rao: I explicitly added this part to categorize the change log into bugs, improvements, incompatible changes etc. rather than a log list of uncategorized git log.

falcon_generate_changelist.py
pallavi-rao: Ok. Will remove.

falcon_generate_changelist.py
```diff
@@ -0,0 +1,148 @@
...
+
+for h in hashes:
+    subject = run_cmd("git log %s -1 --pretty='%%s' | head -1" % h).strip()
+    body = run_cmd("git log %s -1 --pretty='%%b'" % h)
+    committer = run_cmd("git log %s -1 --pretty='%%cn <%%ce>' | head -1" % h).strip()
+    body_lines = body.split("\n")
+
+    match = re.search("\[(\w+)\]" , subject)
+    # Default category
+    category = "BUG"
```
ajayyadava: This information is already in JIRA and we should keep JIRA as the source of truth. Hence I feel this should be fetched from JIRA instead of putting it in commit message.

falcon_generate_changelist.py
```diff
@@ -0,0 +1,148 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# Creates CHANGES.txt from git history.
+# "Inspired by" https://github.com/apache/spark/blob/master/dev/create-release/generate-changelist.py
+#
+# Usage:
+#   First set the new release version and previous release Git tag against which the changelog needs to be generated.
```
ajayyadava: It seems that the script is optimized for generating changes between release tags. It will be nice to document this in the release guide in wiki.

falcon_generate_changelist.py
pallavi-rao: I responded to a similar question on PR 33:
The reason this is added is because of the following reasons:

1. JIRA Type may not directly map to our categorization, for example, backward incompatible changes. Sub-Task can be a "New feature". So, this lets the committer override the default categorization, if required.
2. It is just part of commit message similar to title that also can be retrieved from JIRA. It is just making it available in one place as git log.

falcon_generate_changelist.py
```diff
@@ -0,0 +1,132 @@
...
+    os.remove(TMP_CHANGELIST)
+if os.path.isfile(OLD_CHANGELIST):
+    os.remove(OLD_CHANGELIST)
+
+print "Getting commits between tag %s and HEAD" % PREV_RELEASE_GIT_TAG
+hashes = run_cmd("git log %s..HEAD --pretty='%%h'" % PREV_RELEASE_GIT_TAG).split()
+
+print "Getting details of %s commits" % len(hashes)
+
+for h in hashes:
```
ajayyadava: While some projects may be generating change log like (putting in commit order instead of making different sections) I really prefer the way we generate the change log over the other approach. 

I would also like to add one improvement to it on the lines of Hadoop project. We should put all sub tasks under the parent task as they follow same category, this adds valuable context to the change log and makes it more readable.

falcon_generate_changelist.py
ajayyadava: Thanks for pointing that out, I had missed that update.

I think PR 33 is required only to support this way of generating the change log. I have some issues with this and would like to stick with our original structure of change log (see above comment). 

That solves for issues like backward incompatible (available in JIRA as a tag). This approach is also more flexible as sometimes we realise that certain JIRAs are backward incompatible after we have committed it e.g. during testing.

Also this step is manual and unlike the commit message, author etc. which are by default filled with correct values, this always defaults to "BUG". 

I also find prefixing the issue type in the message as distracting and required only because other projects didn't use JIRA to fetch that information.

falcon_generate_changelist.py
```diff
@@ -0,0 +1,132 @@
...
+    category = "BUG"
+    if match:
+        category = match.group(1).upper()
+        subject = re.sub("\[(\w+)\]", "", subject)
+    file_name = TMP_CHANGELIST + "_" + category
+    if not (os.path.isfile(file_name)):
+        append_to_changelist(file_name, category+"S")
+        append_to_changelist(file_name, "=======================================")
+
+    # Generate GitHub PR URL for easy access if possible
```
ajayyadava: I think since we always add issue number and JIRA contains pull request / patch, we don't require this either.

bvellanki:
@pallavi-rao @ajayyadava  : When will this script be committed? Please let us know if you have ETA

ajayyadava:
@bvellanki @pallavi-rao I propose to use Apache Yetus. It has several benefits over the existing approach

1. Generating changes.txt from jira is much more flexible than from commit messages. 
As I noted earlier also, commits can't be modified, so if a developer forgets to attach metadata / attaches wrong metadata then generating release notes will be incorrect.
2. Yetus provides nice markdown formatted change log with links to each issue. They are easy to read and you can go to the jira by clicking through the github.
3. It can also automate the generation of release notes, if we add them to corresponding JIRA.
4. We don't need to write / maintain and go through all the existing/future learning that YETUS went through.

I have tested it and will send a pull request for the docs generated for all the releases through YETUS.

ajayyadava:
We have started using Apache Yetus and this change is not required anymore. I will be closing this pull request as part of the clean up.

pallavi-rao:
@ajayyadava, agreed. Since we will be using Apache Yetus, this is no longer required. Closing...

PR #35, opened 2016-02-08
FALCON-1783 Fix ProcessUpdateTest and SearchApiTest to use prism

pisaychuk:
We have some weak tests which work in embedded mode, but fail in distributed. Issue exists because in case of distributed mode, entity submit is not allowed via server


pragya-mittal:
+1. LGTM.

PR #38, opened 2016-02-10
FALCON-887 Support for multiple lib paths in falcon process

sowmyaramesh:


oozie/src/main/java/org/apache/falcon/oozie/OozieBundleBuilder.java
```diff
@@ -106,11 +108,24 @@ public Properties build(Cluster cluster, Path buildPath) throws FalconException
...
 
         //Add libpath
-        String libPath = getLibPath(buildPath);
-        if (libPath != null) {
-            properties.put(OozieClient.LIBPATH, getStoragePath(libPath));
+        String path = getLibPath(buildPath);
+        if (StringUtils.isNotBlank(path)) {
+            String storageLibPaths = null;
+            String[] libPaths = path.split(WF_LIB_SEPARATOR);
+            for (String libPath : libPaths) {
```
peeyushb: In the for loop storageLibPaths isNotBlank() is getting multiple times to build a String for storageLibPaths. Shall not be this avoided appropriately.

oozie/src/main/java/org/apache/falcon/oozie/OozieBundleBuilder.java
```diff
@@ -106,11 +108,24 @@ public Properties build(Cluster cluster, Path buildPath) throws FalconException
         properties.setProperty(AbstractWorkflowEngine.NAME_NODE, ClusterHelper.getStorageUrl(cluster));
 
         //Add libpath
-        String libPath = getLibPath(buildPath);
-        if (libPath != null) {
-            properties.put(OozieClient.LIBPATH, getStoragePath(libPath));
+        String path = getLibPath(buildPath);
+        if (StringUtils.isNotBlank(path)) {
+            String storageLibPaths = null;
```
peeyushb: Should StringBuilder be used instead of String to build the String for storageLibPaths.

oozie/src/main/java/org/apache/falcon/oozie/OozieBundleBuilder.java
```diff
@@ -106,11 +108,24 @@ public Properties build(Cluster cluster, Path buildPath) throws FalconException
         properties.setProperty(AbstractWorkflowEngine.NAME_NODE, ClusterHelper.getStorageUrl(cluster));
 
         //Add libpath
-        String libPath = getLibPath(buildPath);
-        if (libPath != null) {
-            properties.put(OozieClient.LIBPATH, getStoragePath(libPath));
+        String path = getLibPath(buildPath);
```
peeyushb: Adding storage LibPath is inside the function Properties build(). I think String storageLibPath should be returned from function and then add the returned string as value to OozieClient.LIBPATH.

oozie/src/main/java/org/apache/falcon/oozie/process/ProcessExecutionWorkflowBuilder.java
```diff
@@ -65,6 +65,8 @@
     private static final Set<String> FALCON_PROCESS_HIVE_ACTIONS = new HashSet<String>(
         Arrays.asList(new String[]{PREPROCESS_ACTION_NAME, USER_ACTION_NAME, }));
 
+    private static final String WF_LIB_SEPARATOR = ",";
```
peeyushb: "WF_LIB_SEPARATOR" has been defined multiple times in the patch, can we define once it in some common utility class and use it across.

pavankumar526:
I think we should document this otherwise user's wont get to know about this. Atleast we should update entity definition adding this

oozie/src/main/java/org/apache/falcon/oozie/OozieBundleBuilder.java
sowmyaramesh: I removed isNotBlank check to add WF_LIB_SEPARATOR. 

Now code is 
storageLibPaths = StringUtils.isNotBlank(storageLibPaths) ?
                            storageLibPaths.concat(EntityUtil.WF_LIB_SEPARATOR).concat(tempPath) : tempPath;

This is required to avoid NPE and also separator should be avoided if there are multiple entries.

oozie/src/main/java/org/apache/falcon/oozie/OozieBundleBuilder.java
sowmyaramesh: Fixed

oozie/src/main/java/org/apache/falcon/oozie/OozieBundleBuilder.java
sowmyaramesh: I don't understand your comment completely. Are you saying getStoragePath should return comma separated lib paths? 

getStoragePath is called from multiple places and for other cases input is not comma spearated list and hence becomes unnecessary processing. Hence I moved the logic only where WF libPaths are processed. You can verify all the callers of getStoragePath.

oozie/src/main/java/org/apache/falcon/oozie/process/ProcessExecutionWorkflowBuilder.java
sowmyaramesh: Done, moved to EntityUtil

oozie/src/main/java/org/apache/falcon/oozie/OozieBundleBuilder.java
peeyushb: I meant to say that this new section of code for adding storage LibPath, into Properties, from line 110-125 is inside the function Properties build(). So I am of the opinion that this section of code for storageLibPath should be defined in a String returned function and then add the returned string OozieClient.LIBPATH.

sowmyaramesh: @peeyushb  Moved that logic to private method.

peeyushb:
Thanks @sowmyaramesh for working on this. It looks good. +1.

peeyushb:
@sowmyaramesh Jenkins build failed. Is the failure related to the patch.

sowmyaramesh:
@peeyushb : From the logs
"Build timed out (after 90 minutes). Marking the build as failed.
Build was aborted"

Not related to my change.

pavankumar526:
+1 Looks good

PR #43, opened 2016-02-12
FALCON-1567 Test case for Lifecycle feature

pragya-mittal:


pisaychuk:
:+1: looks good to me

PR #44, opened 2016-02-15
FALCON-1566 Add test for SLA monitoring API

pragya-mittal:


falcon-regression/merlin/src/test/java/org/apache/falcon/regression/SLA/FeedSLAMonitoringTest.java
```diff
@@ -0,0 +1,278 @@
...
+
+    private static final Comparator<SchedulableEntityInstance> DEPENDENCY_COMPARATOR =
+            new Comparator<SchedulableEntityInstance>() {
+                @Override
+                public int compare(SchedulableEntityInstance o1, SchedulableEntityInstance o2) {
+                    int tagDiff = o1.getTags().compareTo(o2.getTags());
+                    if (tagDiff != 0) {
+                        return tagDiff;
+                    }
+                    int clusterDiff = o1.getCluster().compareTo(o2.getCluster());
```
deepakbarr: I think you can remove lines 86-102 by simply  "return o1.compareTo(o2)".

Also, I 'm not sure why tags are not compared in SchedulableEntityInstance.compareTo() method. If tags should be compared in that SchedulableEntityInstance class, you can remove lines 82-85 as well. What do you think ?

falcon-regression/merlin/src/test/java/org/apache/falcon/regression/SLA/FeedSLAMonitoringTest.java
```diff
@@ -0,0 +1,278 @@
...
+
+        LOGGER.info(response.getMessage());
+
+        validateInstances(response, expectedInstances);
+
+        /**TEST : Create missing dependencies with _SUCCESS directory and check sla response
+         */
+
+        String dateEntry = (String) instanceEntityMap.keySet().toArray()[1];
+        System.out.println(dateEntry + "/" + instanceEntityMap.get(dateEntry));
```
pisaychuk: System.out.println ?

falcon-regression/merlin/src/test/java/org/apache/falcon/regression/SLA/FeedSLAMonitoringTest.java
```diff
@@ -0,0 +1,278 @@
...
+        validateInstances(response, expectedInstances);
+
+        /** TEST : Delete feed and check sla response
+         */
+        String deletedFeed = slaFeedNames.get(0);
+        prism.getFeedHelper().deleteByName(deletedFeed, null);
+
+        for (Map.Entry<String, List<SchedulableEntityInstance>> entry : instanceEntityMap.entrySet())
+        {
+            System.out.println(entry.getKey() + "/" + entry.getValue());
```
pisaychuk: System.out.println ?

falcon-regression/merlin/src/test/java/org/apache/falcon/regression/SLA/FeedSLAMonitoringTest.java
pragya-mittal: I intended to use compareTo but since it tags comparison was missing, overrided it by my method. Anyways I can use compareTo with tags handled explicitly. Will make the changes.

falcon-regression/merlin/src/test/java/org/apache/falcon/regression/SLA/FeedSLAMonitoringTest.java
pragya-mittal: Will remove it and submit new pr.

falcon-regression/merlin/src/test/java/org/apache/falcon/regression/SLA/FeedSLAMonitoringTest.java
pragya-mittal: Will remove it and submit new patch.

falcon-regression/merlin/src/test/java/org/apache/falcon/regression/SLA/FeedSLAMonitoringTest.java
deepakbarr: Not sure why tags are not compared in SchedulableEntityInstance.compareTo(). If it should be there, It's better to add tag comparison to SchedulableEntityInstance.compareTo(). Otherwise, people might end up doing tag comparison explicitly at multiple places when they feel the need. Not a good way.

pragya-mittal: I agree with your point. Created https://github.com/apache/falcon/pull/45 for the same.

pisaychuk:
:+1: looks good!

PR #45, opened 2016-02-17
FALCON-1832 Adding tags in SchedulableEntityInstance.compareTo()

pragya-mittal:


client/src/main/java/org/apache/falcon/resource/SchedulableEntityInstance.java
```diff
@@ -170,6 +170,11 @@ public int compareTo(SchedulableEntityInstance o) {
             return result;
         }
 
+        result = this.tags.compareTo(o.tags);
+        if (result != 0) {
+            return result;
+        }
+
```
deepakbarr: This could potentially throw NPE since tags could be null. Instead, you can use null-safe comparison provided by ObjectUtils class -
result = ObjectUtils.compare(this.tags, o.tags);

ajayyadava: Agree with @deepakbarr @pragya-mittal  can you please make the relevant change and update the pull request?

common/src/main/java/org/apache/falcon/entity/FeedHelper.java
```diff
@@ -437,10 +436,20 @@ public static Date getNextFeedInstanceDate(Date alignedDate, Feed feed) {
         Cluster cluster = getCluster(feed, clusterName);
         if (cluster != null) {
             if (isLifecycleEnabled(feed, clusterName)) {
-                String policy = getRetentionStage(feed, clusterName).getPolicy();
-                policy = StringUtils.isBlank(policy)
-                        ? FeedLifecycleStage.RETENTION.getDefaultPolicyName() : policy;
-                result.add(policy);
+                String policy = "";
```
pallavi-rao: getPolicies method : rather than return the generic list of all policies across stages, please have separate methods. Clubbing policies across lifecycle stages does not make sense.

common/src/main/java/org/apache/falcon/entity/FeedHelper.java
```diff
@@ -1295,4 +1304,17 @@ public static String getPropertyValue(Feed feed, String propName) {
         return storage.getListing(feed, clusterName, locationType, start, end);
     }
 
+    public static String getArchivalPath(Feed feed, String clusterName) throws FalconException {
+        String archivalPath = "";
+        ArchivalStage archivalStage = getArchivalStage(feed, clusterName);
+        if (archivalStage != null) {
+            Location location = archivalStage.getLocation();
+            if ((location != null) && (location.getPath() != null)) {
```
pallavi-rao: Use StringUtils.

common/src/main/java/org/apache/falcon/entity/parser/FeedEntityParser.java
```diff
@@ -157,14 +157,30 @@ private void validateLifecycle(Feed feed) throws FalconException {
         LifecyclePolicyMap map = LifecyclePolicyMap.get();
         for (Cluster cluster : feed.getClusters().getClusters()) {
             if (FeedHelper.isLifecycleEnabled(feed, cluster.getName())) {
-                if (FeedHelper.getRetentionStage(feed, cluster.getName()) == null) {
-                    throw new ValidationException("Retention is a mandatory stage, didn't find it for cluster: "
+                if ((FeedHelper.getRetentionStage(feed, cluster.getName()) != null) || (FeedHelper.getArchivalStage(feed, cluster.getName()) != null)) {
+                    validateRetentionStage(feed, cluster);
+                    validateArchivalStage(feed, cluster);
+                    for (String policyName : FeedHelper.getPolicies(feed, cluster.getName())) {
```
pallavi-rao: Once you split the getPolicies method. Use the appropriate ones inside validateRetentionStage and validateArchivalStage.

common/src/main/java/org/apache/falcon/entity/parser/FeedEntityParser.java
```diff
@@ -157,14 +157,30 @@ private void validateLifecycle(Feed feed) throws FalconException {
...
+    private void validateRetentionStage(Feed feed, Cluster cluster) throws FalconException {
+        if (FeedHelper.getRetentionStage(feed, cluster.getName()) != null) {
+            validateRetentionFrequency(feed, cluster.getName());
+        }
+    }
+
+    private void validateArchivalStage(Feed feed, Cluster cluster) throws FalconException {
+        if (FeedHelper.getArchivalStage(feed, cluster.getName()) != null) {
+            if (FeedHelper.getArchivalStage(feed, cluster.getName()).getLocation().getPath().isEmpty()) {
+                throw new ValidationException("Location path cannot be empty.");
```
pallavi-rao: A little more details on the message like the mention of archival stage for the given cluster will be useful.

common/src/main/java/org/apache/falcon/lifecycle/FeedLifecycleStage.java
```diff
@@ -22,7 +22,8 @@
  */
 public enum FeedLifecycleStage {
```
pallavi-rao: This class needs to be re-visited. A common set of policies for all lifecycles and a default policy across stages does not make sense to me. May be have a base class and child classes.

PR #46, opened 2016-02-18
FALCON-1621 Lifecycle of entity gets missed when prism and falcon server communicates

PraveenAdlakha:


sandeepSamudrala:
+1. looks good
Can you check if something else is also being missed apart from list? date or so?

PraveenAdlakha:
yes have verified date is coming as String so we should be good.

pallavi-rao:
+1. Will merge.

PR #47, opened 2016-02-18
FALCON-1111 Instance update on titan DB based on JMS notifications on workflow jobs

yzheng-hortonworks:
Add running, succeeded, failed, killed and suspended instances to titan DB based on JMS notifications on workflow jobs. Instance-entity edge properties (e.g. nominal time, status) are also added for vertex-centric indexing.


common/src/main/java/org/apache/falcon/metadata/MetadataMappingService.java
```diff
@@ -275,64 +303,49 @@ public Void execute(TransactionalGraph transactionalGraph) throws Exception {
...
+            updateReplicatedFeedInstance(context);
             break;
         case DELETE:
-            onFeedInstanceEvicted(context);
+            updateEvictedFeedInstance(context);
             break;
         case IMPORT:
-            onFeedInstanceImported(context);
+            updateImportedFeedInstance(context);
             break;
```
vramachan: EXPORT operation is missing originally. We need to address that separately via JIRA.

common/src/main/java/org/apache/falcon/metadata/InstanceRelationshipGraphBuilder.java
```diff
@@ -280,20 +271,15 @@ public void addImportedInstance(WorkflowExecutionContext context) throws FalconE
...
-            LOG.info("{} instance vertex {} does not exist, add it",
-                    RelationshipType.FEED_INSTANCE, feedInstanceName);
-            feedInstanceVertex = addFeedInstance(// add a new instance
-                    feedInstanceName, context, feedName, context.getSrcClusterName());
-        }
+        Vertex feedInstanceVertex = addFeedInstance(
+                feedInstanceName, context, feedName, context.getSrcClusterName());
+
+        Map<RelationshipProperty, String> properties = new HashMap<RelationshipProperty, String>();
+        properties.put(RelationshipProperty.TIMESTAMP, context.getTimeStampAsISO8601());
```
vramachan: The edge is populated with nominal time alone and not the status. Let us add status to the edge to be consistent with other entity operations so that the imported instances are searchable.

common/src/main/java/org/apache/falcon/metadata/InstanceRelationshipGraphBuilder.java
```diff
@@ -250,20 +248,13 @@ public void addEvictedInstance(WorkflowExecutionContext context) throws FalconEx
...
-                LOG.info("{} instance vertex {} does not exist, add it",
-                        RelationshipType.FEED_INSTANCE, feedInstanceName);
-                feedInstanceVertex = addFeedInstance(// add a new instance
-                        feedInstanceName, context, feedName, clusterName);
-            }
+            Vertex feedInstanceVertex = addFeedInstance(
+                    feedInstanceName, context, feedName, clusterName);
 
+            Map<RelationshipProperty, String> properties = new HashMap<RelationshipProperty, String>();
+            properties.put(RelationshipProperty.TIMESTAMP, context.getTimeStampAsISO8601());
```
vramachan: The edge has only nominal time. Please add the STATUS also to the edge.

common/src/main/java/org/apache/falcon/metadata/InstanceRelationshipGraphBuilder.java
```diff
@@ -215,19 +219,13 @@ public void addReplicatedInstance(WorkflowExecutionContext context) throws Falco
...
-            LOG.info("{} instance vertex {} does not exist, add it",
-                    RelationshipType.FEED_INSTANCE, feedInstanceName);
-            feedInstanceVertex = addFeedInstance(// add a new instance
-                    feedInstanceName, context, feedName, context.getSrcClusterName());
-        }
+        Vertex feedInstanceVertex = addFeedInstance(
+                feedInstanceName, context, feedName, context.getSrcClusterName());
 
+        Map<RelationshipProperty, String> properties = new HashMap<RelationshipProperty, String>();
+        properties.put(RelationshipProperty.TIMESTAMP, context.getTimeStampAsISO8601());
```
vramachan: The edge has only nominal time. Please add the STATUS also to the edge.

common/src/main/java/org/apache/falcon/metadata/MetadataMappingService.java
yzheng-hortonworks: Thank you, Venky.

common/src/main/java/org/apache/falcon/metadata/InstanceRelationshipGraphBuilder.java
yzheng-hortonworks: The existing property is for message time, which is inherited from original code. I've added nominal time and status to edge properties for search of imported instance. Also added edge properties to search replicated and evicted instances as you commented below. Thanks for pointing this out, Venky.

common/src/main/java/org/apache/falcon/metadata/InstanceRelationshipGraphBuilder.java
```diff
@@ -250,20 +250,15 @@ public void addEvictedInstance(WorkflowExecutionContext context) throws FalconEx
...
-            if (feedInstanceVertex == null) { // No record of instances NOT generated by Falcon
-                LOG.info("{} instance vertex {} does not exist, add it",
-                        RelationshipType.FEED_INSTANCE, feedInstanceName);
-                feedInstanceVertex = addFeedInstance(// add a new instance
-                        feedInstanceName, context, feedName, clusterName);
-            }
+            Vertex feedInstanceVertex = addFeedInstance(
+                    feedInstanceName, context, feedName, clusterName);
 
+            Map<RelationshipProperty, String> properties = new HashMap<RelationshipProperty, String>();
```
bvellanki: minor nit : Lines 225-228 are exactly same, may be this can be a method?

common/src/main/java/org/apache/falcon/metadata/InstanceRelationshipGraphBuilder.java
```diff
@@ -280,20 +275,17 @@ public void addImportedInstance(WorkflowExecutionContext context) throws FalconE
...
-        if (feedInstanceVertex == null) { // No record of instances NOT generated by Falcon
-            LOG.info("{} instance vertex {} does not exist, add it",
-                    RelationshipType.FEED_INSTANCE, feedInstanceName);
-            feedInstanceVertex = addFeedInstance(// add a new instance
-                    feedInstanceName, context, feedName, context.getSrcClusterName());
-        }
+        Vertex feedInstanceVertex = addFeedInstance(
+                feedInstanceName, context, feedName, context.getSrcClusterName());
+
+        Map<RelationshipProperty, String> properties = new HashMap<RelationshipProperty, String>();
```
bvellanki: Code repetition.

bvellanki:
Since FALCON-1838 will address export, +1 for the patch.

vrangan:
+1 Looks good

common/src/main/java/org/apache/falcon/metadata/InstanceRelationshipGraphBuilder.java
yzheng-hortonworks: Created a new function edgePropertiesForIndexing to generate "properties". Note that the property TIMESTAMP is inherited from the original code, but not for indexing, so it is added after calling edgePropertiesForIndexing. Also, edgePropertiesForIndexing is used in other places that doesn't need TIMESTAMP (e.g. addProcessInstance).

bvellanki:
+1 for the patch, will commit shortly

PR #48, opened 2016-02-23
FALCON-1835 Falcon should do coord rerun rather than workflow rerun ...

pallavi-rao:
…to ensure concurrency

Patch contains the following changes:
1. Upgrade to Oozie client 4.2 (from 4.1).
2. Remove workflow rerun completely and switch to coord rerun.
3. Introduced parentId in WorkflowExecutionContext in order to store the coord action of a given workflow. This will be used during retry and late data rerun.
4. In cases where parentId cannot be populated (notification from Post processing action will not contain this), rerun method queries Oozie to retrieve the parent id.
5. The rerun method now returns the actual status after rerun, rather than hardcoded “RUNNING”.
6. In 4.2, Oozie does a Yarn job kill to kill any lingering executions of previous run of a workflow. It uses RMClientProxy to do the same. Since in local mode, RPC cannot happen, had to introduce LocalFalconRPCClientFactory to prevent RPC in local mode.

### DIFF
--- a/common/src/main/java/org/apache/falcon/workflow/WorkflowExecutionArgs.java
+++ b/common/src/main/java/org/apache/falcon/workflow/WorkflowExecutionArgs.java
@@ -55,6 +55,7 @@ public enum WorkflowExecutionArgs {
     STATUS("status", "status of the user workflow isnstance"),
     WF_ENGINE_URL("workflowEngineUrl", "url of workflow engine server, ex:oozie", false),
     USER_SUBFLOW_ID("subflowId", "external id of user workflow", false),
+    PARENT_ID("parentId", "The parent of the current workflow, typically coord action", false),
 
     WF_START_TIME("workflowStartTime", "workflow start time", false),
     WF_END_TIME("workflowEndTime", "workflow end time", false),
@@ -88,7 +89,6 @@ public enum WorkflowExecutionArgs {
     CONTEXT_FILE("contextFile", "wf execution context file path where wf properties are recorded", false),
     CONTEXT_TYPE("contextType", "wf execution context type, pre or post processing", false),
     COUNTERS("counters", "store job counters", false);
-
 
     private final String name;
     private final String description;
--- a/oozie/src/main/java/org/apache/falcon/workflow/engine/OozieWorkflowEngine.java
+++ b/oozie/src/main/java/org/apache/falcon/workflow/engine/OozieWorkflowEngine.java
@@ -97,11 +97,9 @@ public class OozieWorkflowEngine extends AbstractWorkflowEngine {
                 WorkflowJob.Status.FAILED);
     private static final List<WorkflowJob.Status> WF_SUSPEND_PRECOND = Arrays.asList(WorkflowJob.Status.RUNNING);
     private static final List<WorkflowJob.Status> WF_RESUME_PRECOND = Arrays.asList(WorkflowJob.Status.SUSPENDED);
-    private static final List<WorkflowJob.Status> WF_RERUN_PRECOND =
-        Arrays.asList(WorkflowJob.Status.FAILED, WorkflowJob.Status.KILLED, WorkflowJob.Status.SUCCEEDED);
     private static final List<CoordinatorAction.Status> COORD_RERUN_PRECOND =
-        Arrays.asList(CoordinatorAction.Status.TIMEDOUT, CoordinatorAction.Status.FAILED);
-
+        Arrays.asList(CoordinatorAction.Status.TIMEDOUT, CoordinatorAction.Status.FAILED,
+                CoordinatorAction.Status.KILLED, CoordinatorAction.Status.SUCCEEDED);
     private static final List<Job.Status> BUNDLE_ACTIVE_STATUS =
         Arrays.asList(Job.Status.PREP, Job.Status.RUNNING, Job.Status.SUSPENDED, Job.Status.PREPSUSPENDED,
             Job.Status.RUNNINGWITHERROR, Job.Status.PAUSED, Status.PREPPAUSED, Status.PAUSEDWITHERROR);
@@ -937,17 +935,10 @@ public class OozieWorkflowEngine extends AbstractWorkflowEngine {
             break;
 
         case RERUN:
-            if (jobInfo == null && COORD_RERUN_PRECOND.contains(coordinatorAction.getStatus())) {
-                //Coord action re-run
-                reRunCoordAction(cluster, coordinatorAction);
-                status = Status.RUNNING.name();
-            } else if (jobInfo != null && WF_RERUN_PRECOND.contains(jobInfo.getStatus())) {
-                //wf re-run
-                reRun(cluster, jobInfo.getId(), props, isForced);
-                status = Status.RUNNING.name();
+            if (COORD_RERUN_PRECOND.contains(coordinatorAction.getStatus())) {
+                status = reRunCoordAction(cluster, coordinatorAction, props, isForced).name();
             }
             break;
-
 
         case STATUS:
             if (StringUtils.isNotEmpty(coordinatorAction.getExternalId())) {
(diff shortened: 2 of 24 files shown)


oozie/src/main/java/org/apache/falcon/workflow/engine/OozieWorkflowEngine.java
```diff
@@ -97,11 +97,9 @@
                 WorkflowJob.Status.FAILED);
     private static final List<WorkflowJob.Status> WF_SUSPEND_PRECOND = Arrays.asList(WorkflowJob.Status.RUNNING);
     private static final List<WorkflowJob.Status> WF_RESUME_PRECOND = Arrays.asList(WorkflowJob.Status.SUSPENDED);
-    private static final List<WorkflowJob.Status> WF_RERUN_PRECOND =
-        Arrays.asList(WorkflowJob.Status.FAILED, WorkflowJob.Status.KILLED, WorkflowJob.Status.SUCCEEDED);
     private static final List<CoordinatorAction.Status> COORD_RERUN_PRECOND =
-        Arrays.asList(CoordinatorAction.Status.TIMEDOUT, CoordinatorAction.Status.FAILED);
-
+        Arrays.asList(CoordinatorAction.Status.TIMEDOUT, CoordinatorAction.Status.FAILED,
+                CoordinatorAction.Status.KILLED, CoordinatorAction.Status.SUCCEEDED);
```
sandeepSamudrala: suspended too?

oozie/src/main/java/org/apache/falcon/workflow/engine/OozieWorkflowEngine.java
```diff
@@ -974,30 +965,61 @@ private void performAction(String cluster, JobAction action, CoordinatorAction c
...
+                props = new Properties();
+            }
+            // In case if both props exists one should be removed otherwise it will fail.
+            // This case will occur when user runs workflow with skip-nodes property and
+            // try to do force rerun or rerun with fail-nodes property.
+            if (props.containsKey(OozieClient.RERUN_FAIL_NODES)
+                    && props.containsKey(OozieClient.RERUN_SKIP_NODES)) {
+                LOG.warn("Both " + OozieClient.RERUN_SKIP_NODES + " and " + OozieClient.RERUN_FAIL_NODES
+                        + " are present in workflow params removing" + OozieClient.RERUN_SKIP_NODES);
+                props.remove(OozieClient.RERUN_SKIP_NODES);
```
sandeepSamudrala: I understand that oozie says not to set both properties together, But rather than removing the property, should we not throw a exception and notify the user? Else it might always go unintended by the user.

oozie/src/main/java/org/apache/falcon/workflow/engine/OozieWorkflowEngine.java
pallavi-rao: Right. Missed that. Will add.

oozie/src/main/java/org/apache/falcon/workflow/engine/OozieWorkflowEngine.java
```diff
@@ -974,30 +965,61 @@ private void performAction(String cluster, JobAction action, CoordinatorAction c
...
+                isForced = false;
+            }
+
+            Properties jobprops;
+            // Get conf when workflow is launched.
+            if (coordinatorAction.getExternalId() != null) {
+                WorkflowJob jobInfo = client.getJobInfo(coordinatorAction.getExternalId());
+
+                jobprops = OozieUtils.toProperties(jobInfo.getConf());
+                jobprops.putAll(props);
```
sandeepSamudrala: Wf props being replaced, if lib path gets updated by the time rerun has happened(during a deploy or so), the jars registered in the libpath might not be present by now. Correct me if its not the way.

oozie/src/main/java/org/apache/falcon/workflow/engine/OozieWorkflowEngine.java
pallavi-rao: Hmm.. I remember asking the same question to Pavan when he added this. Unfortunately I can't recollect his answer. Pavan, can you jog my memory as to why it is this way.

oozie/src/main/java/org/apache/falcon/workflow/engine/OozieWorkflowEngine.java
pallavi-rao: wfProps is retrieved from jobInfo and this is how it has been. I believe the idea is that the rerun should happen with same conf as the original run. 

Yes, if some jars have been moved around, it will fail. This is better than rerunning with the new libs when the user didn't mean to use the new ones. If the user really intends to change props, he can always supply them while calling rerun.

oozie/src/main/java/org/apache/falcon/workflow/engine/OozieWorkflowEngine.java
sriksun: Can suspended be re-run or only resumed ?

pallavi-rao: Per documentation, even suspended can be rerun - http://oozie.apache.org/docs/4.2.0/DG_CoordinatorRerun.html. I haven't tested it though. Will try and update.

oozie/src/main/java/org/apache/falcon/workflow/engine/OozieWorkflowEngine.java
sriksun: Agree with @sandeepSamudrala. It is possible we might inadvertently do the opposite of what the user sought to do.

oozie/src/main/java/org/apache/falcon/workflow/engine/OozieWorkflowEngine.java
pallavi-rao: Checked. SUSPENDED actions don't get rerun. Rerun coordinator action must be in TIMEDOUT/SUCCEEDED/KILLED/FAILED. 

"Rerun a SUSPENDED coordinator job, the status will reset to RUNNING." - Means even if a coord job is suspended, when some actions are rerun, the job (and not action) moves to RUNNING. I saw this happen. Same happens to DONEWITHERROR (moves to RUNNING when any action is rerun)

So, won't add SUSPENDED.

rerun/src/test/java/org/apache/falcon/rerun/queue/InMemoryQueueTest.java
```diff
@@ -45,7 +45,7 @@ private void runTest() throws InterruptedException, FalconException {
             Thread.sleep(30);
             long time = System.currentTimeMillis();
             int delay = ((5 - index) / 2) * 50;
-            MyEvent event = new MyEvent("someCluster", Integer.toString(index),
+            MyEvent event = new MyEvent("someCluster", Integer.toString(index), "parent",
```
sandeepSamudrala: should not "parent" string contain "-C@"?

pallavi-rao: Doesn't matter for this particular test as OozieWorkflowEngine API is not invoked.

oozie/src/main/java/org/apache/falcon/workflow/engine/OozieWorkflowEngine.java
sandeepSamudrala: Agreed. Just wondering, if we would send a api response status to succeeded in case of suspended coord action being rerun. I think this has been since long. We can track it in another Jira.

oozie/src/main/java/org/apache/falcon/workflow/engine/OozieWorkflowEngine.java
pavankumar526: Suppose a user do force re-run once then RERUN_FAIL_NODES will be set  and he again doing rerun by setting other property RERUN_SKIP_NODES, it leads to rerun in failure, that's why one property was removed to make rerun successful. Or vice versa will happen first user rerun by passing some props and he did force rerun it might not work. Thats why we are removing one

common/src/main/java/org/apache/falcon/workflow/WorkflowExecutionArgs.java
```diff
@@ -55,6 +55,7 @@
     STATUS("status", "status of the user workflow isnstance"),
     WF_ENGINE_URL("workflowEngineUrl", "url of workflow engine server, ex:oozie", false),
     USER_SUBFLOW_ID("subflowId", "external id of user workflow", false),
+    PARENT_ID("parentId", "The parent of the current workflow, typically coord action", false),
```
sriksun: Does it make sense for you to get the parent from workflow bean via oozie client on the re-run & late-re-run path instead of bouncing it of notification ?

oozie/src/main/java/org/apache/falcon/workflow/engine/OozieWorkflowEngine.java
```diff
@@ -974,30 +965,61 @@ private void performAction(String cluster, JobAction action, CoordinatorAction c
...
+                                                      Properties props, boolean isForced) throws FalconException {
         try {
             OozieClient client = OozieClientFactory.get(cluster);
+            if (props == null) {
+                props = new Properties();
+            }
+            // In case if both props exists one should be removed otherwise it will fail.
+            // This case will occur when user runs workflow with skip-nodes property and
+            // try to do force rerun or rerun with fail-nodes property.
+            if (props.containsKey(OozieClient.RERUN_FAIL_NODES)
```
pavankumar526: @pallavi-rao  this check has to be done after copying user props with workflow props not before.

pallavi-rao: You are right. Will move. Since coord rerun doesn't honor these, couldn't test it.

oozie/src/main/java/org/apache/falcon/workflow/engine/OozieWorkflowEngine.java
pavankumar526: @sriksun and @sandeepSamudrala  two properties check should be done after user props copied with workflow props then only controversies can come.

common/src/main/java/org/apache/falcon/workflow/WorkflowExecutionArgs.java
pallavi-rao: Saving an Oozie call since Oozie notification already contains it :-)

oozie/src/main/java/org/apache/falcon/workflow/engine/OozieWorkflowEngine.java
pallavi-rao: Agree on both counts. An error is better than us keeping in sync with Oozie's preconditions. But, that will change the current behavior and will need update across multiple APIs to keep them consistent. Would rather handle that in a separate JIRA.

oozie/src/main/java/org/apache/falcon/workflow/engine/OozieWorkflowEngine.java
```diff
@@ -974,30 +965,61 @@ private void performAction(String cluster, JobAction action, CoordinatorAction c
...
+
             client.reRunCoord(coordinatorAction.getJobId(), RestConstants.JOB_COORD_SCOPE_ACTION,
-                Integer.toString(coordinatorAction.getActionNumber()), true, true);
-            assertCoordActionStatus(cluster, coordinatorAction.getId(),
-                org.apache.oozie.client.CoordinatorAction.Status.RUNNING,
-                org.apache.oozie.client.CoordinatorAction.Status.WAITING,
-                org.apache.oozie.client.CoordinatorAction.Status.READY);
+                    Integer.toString(coordinatorAction.getActionNumber()), true, true, !isForced, jobprops);
             LOG.info("Rerun job {} on cluster {}", coordinatorAction.getId(), cluster);
+            return assertCoordActionStatus(cluster, coordinatorAction.getId(),
```
pavankumar526: I don't think its possible what if coord-action succedded so fast ?

pallavi-rao: Very very unlikely as this is a synchronous call. But, I can add that anyway.

pavankumar526: But if we add the whole point of assertion will go away right, like we want to rerun the succeeded instance. I think it's ok we can leave like this.

oozie/src/main/java/org/apache/falcon/workflow/engine/OozieWorkflowEngine.java
sandeepSamudrala: I agree that check needs to be done. but my question is should we not throw an exception rather than removing ?

sandeepSamudrala:
+1.  looks good..

oozie/src/main/java/org/apache/falcon/workflow/engine/OozieWorkflowEngine.java
```diff
@@ -974,30 +965,61 @@ private void performAction(String cluster, JobAction action, CoordinatorAction c
...
+                if (props.containsKey(OozieClient.RERUN_FAIL_NODES)
+                        && props.containsKey(OozieClient.RERUN_SKIP_NODES)) {
+                    String msg = "Both " + OozieClient.RERUN_SKIP_NODES + " and " + OozieClient.RERUN_FAIL_NODES
+                            + " are present in workflow params for " + coordinatorAction.getExternalId();
+                    LOG.error(msg);
+                    throw new FalconException(msg);
+                }
+
+                //if user has set any of these oozie rerun properties then force rerun flag is ignored
+                if (props.containsKey(OozieClient.RERUN_FAIL_NODES)) {
```
pavankumar526: This condition might have problem, if user rerun with this property, he cannot do force rerun again. Force flag will be always set false.

pavankumar526:
+1 looks good.

pallavi-rao:
Thanks everyone for the review. Will merge this shortly.

PR #49, opened 2016-02-24
FALCON-1839 Test case for APIs for entities scheduled on native scheduler

pragya-mittal:


falcon-regression/merlin-core/src/main/java/org/apache/falcon/regression/core/util/NativeInstanceUtil.java
```diff
@@ -0,0 +1,176 @@
...
+        LOGGER.info(String.format("Sleep for %d seconds", sleepTime));
+        for (int i = 0; i < maxTries; i++) {
+            InstancesResult statusResult = cluster.getProcessHelper().getProcessInstanceStatus(entityName, params);
+            if (statusResult.getInstances() != null) {
+                if (statusResult.getInstances()[0].getStatus().name().equals(expectedStatus.toString())) {
+                    return;
+                }
+            }
+            LOGGER.info(type + " " + entityName + " still doesn't have expected status");
+            TimeUtil.sleepSeconds(5);
```
deepakbarr: TimeUtil.sleepSeconds(5) ?? 
I think you intended TimeUtil.sleepSeconds(sleepTime) ??

falcon-regression/merlin/src/test/java/org/apache/falcon/regression/nativeScheduler/NativeSchedulerAPITest.java
```diff
@@ -0,0 +1,306 @@
...
+/**
+ * Schedule process via native scheduler and test available APIs.
+ */
+
+@Test(groups = {"distributed", "embedded" })
+public class NativeSchedulerAPITest extends BaseTestClass {
+    private ColoHelper cluster1 = servers.get(0);
+    private String baseTestHDFSDir = cleanAndGetTestDir();
+    private String aggregateWorkflowDir = baseTestHDFSDir + "/aggregator";
+    private static final Logger LOGGER = Logger.getLogger(NativeScheduleTest.class);
```
deepakbarr: Logger.getLogger(NativeScheduleTest.class) ?? Wrong class name?

falcon-regression/merlin-core/src/main/java/org/apache/falcon/regression/core/util/NativeInstanceUtil.java
```diff
@@ -0,0 +1,176 @@
...
+     * @param cluster     ColoHelper - colo on which API to be executed
+     * @param entity      definition of entity which describes job
+     * @param startTime   start time of instance
+     * @param endTime     end time of instance
+     */
+    public static void waitTillInstancesAreCreated(ColoHelper cluster, Entity entity, String startTime, String endTime)
+        throws InterruptedException, IOException, AuthenticationException,
+            URISyntaxException {
+        int sleep = INSTANCES_CREATED_TIMEOUT * 60 / 5;
+        waitTillInstancesAreCreated(cluster, entity, startTime, endTime, sleep);
```
deepakbarr: INSTANCES_CREATED_TIMEOUT * 60 / 5 - This does not seem intuitive. Why divide by 5 ?

falcon-regression/merlin-core/src/main/java/org/apache/falcon/regression/core/util/NativeInstanceUtil.java
```diff
@@ -0,0 +1,176 @@
...
+     */
+    public static void waitTillInstancesAreCreated(ColoHelper cluster, Entity entity, String startTime,
+        String endTime, int totalMinutesToWait)
+        throws URISyntaxException, AuthenticationException, InterruptedException, IOException {
+        String entityName = entity.getName();
+        EntityType type = entity.getEntityType();
+        String params = "?start=" + startTime;
+        params += (endTime.isEmpty() ? "" : "&end=" + endTime);
+
+        for (int sleepCount = 0; sleepCount < totalMinutesToWait; sleepCount++) {
```
deepakbarr: I assume the loop keeps checking the status for "totalMinutesToWait" minutes. In the end, TimeUtil.sleepSeconds(5) should ideally be TimeUtil.sleepSeconds(60) ?

falcon-regression/merlin/src/test/java/org/apache/falcon/regression/nativeScheduler/NativeSchedulerAPITest.java
pragya-mittal: Will fix

falcon-regression/merlin-core/src/main/java/org/apache/falcon/regression/core/util/NativeInstanceUtil.java
```diff
@@ -0,0 +1,176 @@
...
+        String params = "?start=" + instanceTime + "&end=" + endTime;
+
+        int maxTries = 50;
+        int totalSleepTime = totalMinutesToWait * 60;
+        int sleepTime = totalSleepTime / maxTries;
+        LOGGER.info(String.format("Sleep for %d seconds", sleepTime));
+        for (int i = 0; i < maxTries; i++) {
+            InstancesResult statusResult = cluster.getProcessHelper().getProcessInstanceStatus(entityName, params);
+            if (statusResult.getInstances() != null) {
+                if (statusResult.getInstances()[0].getStatus().name().equals(expectedStatus.toString())) {
```
pisaychuk: we just can compare enums themselves using ==

falcon-regression/merlin-core/src/main/java/org/apache/falcon/regression/core/util/NativeInstanceUtil.java
```diff
@@ -0,0 +1,176 @@
...
+     * @param entity      definition of entity which describes job
+     * @param startTime   start time of instance
+     * @param endTime     end time of instance
+     * @param totalMinutesToWait     total time(in minutes) to wait for instance creation
+     */
+    public static void waitTillInstancesAreCreated(ColoHelper cluster, Entity entity, String startTime,
+        String endTime, int totalMinutesToWait)
+        throws URISyntaxException, AuthenticationException, InterruptedException, IOException {
+        String entityName = entity.getName();
+        EntityType type = entity.getEntityType();
```
pisaychuk: we use entity type once, so we can omit the creation of variable

falcon-regression/merlin-core/src/main/java/org/apache/falcon/regression/core/util/NativeInstanceUtil.java
```diff
@@ -0,0 +1,176 @@
...
+     * @param cluster     ColoHelper - colo on which API to be executed
+     * @param entity      definition of entity which describes job
+     * @param startTime   start time of instance
+     * @param endTime     end time of instance
+     */
+    public static void waitTillInstancesAreCreated(ColoHelper cluster, Entity entity, String startTime, String endTime)
+        throws InterruptedException, IOException, AuthenticationException,
+            URISyntaxException {
+        int sleep = INSTANCES_CREATED_TIMEOUT;
+        waitTillInstancesAreCreated(cluster, entity, startTime, endTime, sleep);
```
pisaychuk: can't we pass INSTANCES_CREATED_TIMEOUT directly ?

falcon-regression/merlin-core/src/main/java/org/apache/falcon/regression/core/util/NativeInstanceUtil.java
```diff
@@ -0,0 +1,176 @@
...
+import org.apache.hadoop.security.authentication.client.AuthenticationException;
+import org.apache.log4j.Logger;
+import org.apache.oozie.client.CoordinatorAction;
+import org.testng.Assert;
+
+import java.io.IOException;
+import java.net.URISyntaxException;
+
+/**
+ * util functions related to instanceTest.
```
pisaychuk: please clarify what exact test it relates to. Or describe general purpose of methods (it might be reused by another tests in future).

falcon-regression/merlin-core/src/main/java/org/apache/falcon/regression/core/util/NativeInstanceUtil.java
```diff
@@ -0,0 +1,176 @@
...
+     * @param expectedStatus    expected status we are waiting for
+     * @param frequency         frequency of process/feed
+     * @param totalMinutesToWait     total time(in minutes) to wait
+     */
+    public static void waitTillInstanceReachState(ColoHelper cluster, Entity entity, String instanceTime,
+        CoordinatorAction.Status expectedStatus, Frequency frequency, int totalMinutesToWait)
+        throws URISyntaxException, AuthenticationException, InterruptedException, IOException {
+        String entityName = entity.getName();
+        EntityType type = entity.getEntityType();
+
```
pisaychuk: We can omit 'type' variable as it is used only once.

falcon-regression/merlin/src/test/java/org/apache/falcon/regression/nativeScheduler/NativeSchedulerAPITest.java
```diff
@@ -0,0 +1,306 @@
...
+import org.testng.annotations.AfterMethod;
+import org.testng.annotations.BeforeClass;
+import org.testng.annotations.BeforeMethod;
+import org.testng.annotations.Test;
+
+/**
+ * Schedule process via native scheduler and test available APIs.
+ */
+
+@Test(groups = {"distributed", "embedded" })
```
pisaychuk: we can omit distributed as we consider embedded tests runnable in distributed setup also (while distributed runnable only in distributed setup)

falcon-regression/merlin/src/test/java/org/apache/falcon/regression/nativeScheduler/NativeSchedulerAPITest.java
```diff
@@ -0,0 +1,306 @@
...
+                .getEntitySummary(processMerlin.getClusterNames().get(0), "filterBy=STATUS:RUNNING")
+                .getEntitySummaryResult().getEntitySummaries();
+        Assert.assertEquals(summaries.length, 1, "There should be one RUNNING process filtered.");
+        Assert.assertEquals(summaries[0].getName(), processMerlin.getName(),
+                "Summary shows invalid suspended process.");
+    }
+
+    /*Test for instance dependency, list, params, logs, running and triage API and check that it is successful.
+     */
+    @Test(enabled = true)
```
pisaychuk: let's clean up (enabled = true) as it is default behaviour

falcon-regression/merlin/src/test/java/org/apache/falcon/regression/nativeScheduler/NativeSchedulerAPITest.java
```diff
@@ -0,0 +1,306 @@
...
+        AssertUtil.assertSucceeded(cluster1.getProcessHelper().resume(processMerlin.toString()));
+
+        Assert.assertTrue(Util.parseResponse(cluster1.getProcessHelper()
+                .getStatus(processMerlin.toString())).getMessage()
+                .contains("RUNNING"));
+    }
+
+    /*Test for entity definition, list, dependency, update and summary API and check that it is successful.
+     */
+    @Test(enabled = true)
```
pisaychuk: let's clean up (enabled = true) as it is default behaviour

pisaychuk:
@pragya-mittal I have left minor comments. Everything else looks good to me.

falcon-regression/merlin-core/src/main/java/org/apache/falcon/regression/core/util/NativeInstanceUtil.java
pragya-mittal: @deepakbarr  I have made it to 60 seconds. But IMO waiting for a minute in every cycle for instance to get materialize is not very intuitive. Here is the thing, in oozie there is materialization service running which wakes up every 300 secs(default value) and materializes instances. But this is not the case in scheduler. In native scheduler if instanceTime < now, instance should get created. I thought of putting the basic function and depending on future observation and tests I will modify.If you have any other approach , do share!

pragya-mittal: murali.msse@gmail.com Thoughts please?

muraliramasami: @pragya-mittal  IMO,60secs looks long. Because it may slow down our test execution. We can try with 5-10secs. Based on the future observation we can increase if it is required.

pragya-mittal: @muraliramasami Changing it to 10 secs.

falcon-regression/merlin-core/src/main/java/org/apache/falcon/regression/core/util/NativeInstanceUtil.java
pragya-mittal: Done.

falcon-regression/merlin-core/src/main/java/org/apache/falcon/regression/core/util/NativeInstanceUtil.java
pragya-mittal: Yes we can. Passing it directly now.

falcon-regression/merlin-core/src/main/java/org/apache/falcon/regression/core/util/NativeInstanceUtil.java
pragya-mittal: Will add more docs.

falcon-regression/merlin/src/test/java/org/apache/falcon/regression/nativeScheduler/NativeSchedulerAPITest.java
pragya-mittal: @pisaychuk @muraliramasami As going by our offline discussion, I am keeping both groups so as to ease nightly regression.

falcon-regression/merlin/src/test/java/org/apache/falcon/regression/nativeScheduler/NativeSchedulerAPITest.java
pragya-mittal: Done.

falcon-regression/merlin/src/test/java/org/apache/falcon/regression/nativeScheduler/NativeSchedulerAPITest.java
pragya-mittal: Done.

pragya-mittal:
@muraliramasami  Addressed review comments. Please check.

muraliramasami:
@pragya-mittal +1 Patch looks good to me.

falcon-regression/merlin-core/src/main/java/org/apache/falcon/regression/core/util/NativeInstanceUtil.java
```diff
@@ -0,0 +1,181 @@
...
+     * @param totalMinutesToWait     total time(in minutes) to wait for instance creation
+     */
+    public static void waitTillInstancesAreCreated(ColoHelper cluster, Entity entity, String startTime,
+        String endTime, int totalMinutesToWait)
+        throws URISyntaxException, AuthenticationException, InterruptedException, IOException {
+        String entityName = entity.getName();
+        EntityType type = entity.getEntityType();
+        String params = "?start=" + startTime;
+        params += (endTime.isEmpty() ? "" : "&end=" + endTime);
+        totalMinutesToWait = (totalMinutesToWait*60)/10;
```
deepakbarr: Not sure about the reason for
 totalMinutesToWait = (totalMinutesToWait*60)/10

pragya-mittal: Currently we are passing 10 minutes which will wait for 10seconds. i.e 100 seconds < 2minutes. But if we pass 10*60/10 = 60 * 100 i.e 600 seconds = 10 minutes. So for every 10 seconds we will check for completion up to 10minutes.

deepakbarr: I think I get what you are trying to do. But this does not seem intuitive. Since you are overwriting totalMinutesToWait variable (totalMinutesToWait = (totalMinutesToWait*60)/10) and now, the name totalMinutesToWait does not actually represent the total minutes wait will happen. How about something like this - 

totalSecondsToWait = totalMinutesToWait * 60;
for (int sleepSeconds = 0; sleepSeconds < totalSecondsToWait; sleepSeconds = sleepSeconds+10) {
// Check instance
 TimeUtil.sleepSeconds(10);
}

deepakbarr: OR  you can use the same logic you used in lines 132-145 in waitTillInstanceReachState() in the same class (using maxTries variable).
I just want to make sure variable names are intuitive.

pragya-mittal:
@muraliramasami  @deepakbarr  Modified NativeInstanceUtil.waitTillInstancesAreCreated based on review comments.

falcon-regression/merlin/src/test/java/org/apache/falcon/regression/nativeScheduler/NativeSchedulerAPITest.java
```diff
@@ -0,0 +1,306 @@
...
+
+        Assert.assertTrue(Util.parseResponse(cluster1.getProcessHelper()
+                .getStatus(processMerlin.toString())).getMessage()
+                .contains("RUNNING"));
+    }
+
+    /*Test for entity definition, list, dependency, update and summary API and check that it is successful.
+     */
+    @Test
+    public void entityDefinitionListDependencyUpdateSummaryTest() throws Exception {
```
pavankumar526: Can we separate out update and other definition checks

pavankumar526: After update we should also check whether new staging path got created or not we have to check.

pragya-mittal: These are basic sanity test cases to see whether APIs are working fine with time based scheduling. I intend to add all the specific cases once data based native scheduling comes in picture. Does it make sense ?

pavankumar526: Its good to add test for delete api also

pragya-mittal: Ideally removeTestClassEntities() deletes all the entities and throws exception if delete fails. That is why didnt add the test for the same. I don't see the need of explicitly adding delete test case.

pavankumar526: Ok agreed. Please do all  checks when databased support comes

pavankumar526:
+1 looks good

pragya-mittal:
@muraliramasami @deepakbarr @pavankumar526  I have addressed all the review comments. Modified  NativeInstanceUtil.waitTillInstancesAreCreated logic (as deepak suggested) and added documentation for the same. Please check once.

falcon-regression/merlin-core/src/main/java/org/apache/falcon/regression/core/util/NativeInstanceUtil.java
```diff
@@ -0,0 +1,186 @@
...
+     * Waits till instance of specific entity will be created during timeout.
+     * Timeout is common for most of usual test cases.
+     * Using status API to retrieve instance information.
+     *
+     * @param cluster     ColoHelper - colo on which API to be executed
+     * @param entity      definition of entity which describes job
+     * @param startTime   start time of instance
+     * @param endTime     end time of instance
+     */
+    public static void waitTillInstancesAreCreated(ColoHelper cluster, Entity entity, String startTime, String endTime)
```
deepakbarr: When is this method called ?? I could not find any usage.

falcon-regression/merlin-core/src/main/java/org/apache/falcon/regression/core/util/NativeInstanceUtil.java
```diff
@@ -0,0 +1,186 @@
...
+     * In other cases use waitTillInstancesAreCreated(ColoHelper,Entity,String,String)
+     * Using status API to retrieve instance information.
+     *
+     * @param cluster     ColoHelper - colo on which API to be executed
+     * @param entity      definition of entity which describes job
+     * @param startTime   start time of instance
+     * @param endTime     end time of instance
+     * @param totalMinutesToWait     total time(in minutes) to wait for instance creation
+     */
+    public static void waitTillInstancesAreCreated(ColoHelper cluster, Entity entity, String startTime,
```
deepakbarr: Is there an assertion missing in this method. Whether instances are created are not, the method behaviour is same. Am I missing something ?

pragya-mittal: Added assertion.

falcon-regression/merlin-core/src/main/java/org/apache/falcon/regression/core/util/NativeInstanceUtil.java
pragya-mittal: I intended to use it but then didnt use it. Anyhow with data based scheduling coming into picture it will be used heavily since we need to check instance availability before going further.

falcon-regression/merlin-core/src/main/java/org/apache/falcon/regression/core/util/NativeInstanceUtil.java
```diff
@@ -0,0 +1,186 @@
...
+        EntityType type = entity.getEntityType();
+
+        String endTime=getNextInstanceTime(instanceTime, frequency);
+        String params = "?start=" + instanceTime + "&end=" + endTime;
+
+        int maxTries = 50;
+        int totalSleepTime = totalMinutesToWait * 60;
+        int sleepTime = totalSleepTime / maxTries;
+        LOGGER.info(String.format("Sleep for %d seconds", sleepTime));
+        for (int i = 0; i < maxTries; i++) {
```
deepakbarr: Here, we have fixed the number of tries to 50 and sleep time because consecutive checks is computed. So, If totalMinutesToWait is 10, then, instance checks with happen every 12 seconds for 50 times. If totalMinutesToWait is 1, then, instance check will happen every 1.2 seconds for 50 times.

IMO, We should have sleepTime as fixed ( say 10 seconds ) and maxTries should be computed according to totalMinutesToWait. If totalMinutesToWait is 1 minute, We will check instance every 10 seconds
 for 6 times total. Similarly, If totalMinutesToWait is 5 minute, We will again check instance every 10 seconds for 30 times total.  What do you think ?

pragya-mittal: I get it you want to implement it the same way as waitTillInstancesAreCreated(). But this approach is also fine, and I don't see any specific advantage of changing the logic here. @muraliramasami Thoughts please ?

pragya-mittal: The same logic is used in InstanceUtil.waitTillInstanceReachState() and it serves the required purpose.

deepakbarr: In InstanceUtil.waitTillInstanceReachState(), both sleep time and retries are fixed. 
I agree that the code will work as it and serves the purpose. It's more about the reusability of waitTillInstanceReachState() method. If someone wants to use this method  and sends  parameter totalMinutesToWait = 1, The instance check will happen every 1.2 seconds which is probably not desired/intended.

muraliramasami: @pragya-mittal Looks like current function serves the purpose but not generic.That is the only concern. I think we can modify the code like:
int totalSleepTime = totalMinutesToWait * 60;
int sleepTime = 10;
int maxTries = totalSleepTime / sleepTime;

@deepakbarr Can you let us know is the above approach is good? Can you share your thoughts?

pragya-mittal: Making it generic now :)

pragya-mittal:
@muraliramasami @deepakbarr Please review. I want to commit this patch so that I can confirm native scheduler behaviour in my nightly regression.

falcon-regression/merlin-core/src/main/java/org/apache/falcon/regression/core/util/NativeInstanceUtil.java
```diff
@@ -0,0 +1,184 @@
...
+        throws URISyntaxException, AuthenticationException, InterruptedException, IOException {
+        String entityName = entity.getName();
+        EntityType type = entity.getEntityType();
+
+        String endTime=getNextInstanceTime(instanceTime, frequency);
+        String params = "?start=" + instanceTime + "&end=" + endTime;
+
+        int totalSleepTime = totalMinutesToWait * 60;
+        int sleepTime = 10;
+        for (int i = 0; i < totalSleepTime; i = i+10) {
```
deepakbarr: instead of i=i+10, make it i=i+sleepTime. Rest looks good.

deepakbarr:
One minor comment. Rest looks good.

falcon-regression/merlin-core/src/main/java/org/apache/falcon/regression/core/util/NativeInstanceUtil.java
pragya-mittal: I was expecting it. Will make the change :)

pragya-mittal:
@deepakbarr Please check now.

deepakbarr:
+1

pragya-mittal:
Will merge the PR tomorrow unless there are more review comments.

PR #50, opened 2016-02-25
FALCON-1724 Falcon CLI.twiki in docs folder is not pointed by index page

PraveenAdlakha:
Currentlt Faclon CLI.twiki is not pointed by any page we need to remove it to avoid the confusion.


pallavi-rao:
@PraveenAdlakha, a quick search shows that Recipes.twiki has some reference to old FalconCLI. Can you please update that too?
```
$ find . -name "*.twiki" | xargs grep -r FalconCLI | grep -v FalconCLI.twiki 
./src/site/twiki/index.twiki:[[falconcli/FalconCLI][Falcon CLI]] implements [[restapi/ResourceList][Falcon's RESTful API]] and
./src/site/twiki/Recipes.twiki:<a href="./FalconCLI.html">Recipe command usage is defined here.</a>
./src/site/twiki/Recipes.twiki:   * Invokes a Tool to substitute the properties in the templated process for the recipe. By default invokes base tool if tool option is not passed. Tool is responsible for generating process entity at the path specified by FalconCLI
./src/site/twiki/Recipes.twiki:   * Looks for the patten ##[A-Za-z0-9_.]*## in the templated process and substitutes it with the properties. Process entity generated after the substitution is written to the empty file passed by FalconCLI
```

pallavi-rao:
+1. Will merge.

PR #57, opened 2016-03-01
FALCON-1802 Data based Support for Native Scheduler

pavankumar526:


common/src/main/java/org/apache/falcon/util/DateUtil.java
```diff
@@ -99,4 +136,142 @@ public static Date now() {
...
+     */
+    public static TimeZone getTimeZone(String tzId) {
+        if (tzId == null) {
+            throw new IllegalArgumentException("Timezone cannot be null");
+        }
+        tzId = handleGMTOffsetTZNames(tzId);    // account for GMT-####
+        TimeZone tz = TimeZone.getTimeZone(tzId);
+        // If these are not equal, it means that the tzId is not valid (invalid tzId's return GMT)
+        if (!tz.getID().equals(tzId)) {
+            throw new IllegalArgumentException("Invalid TimeZone: " + tzId);
```
sandeepSamudrala: will not this validation happen at the time of entity submit itself. Will this not be required to check again during schedule?

common/src/main/java/org/apache/falcon/util/TimeUnit.java
```diff
@@ -0,0 +1,39 @@
...
+ */
+
+package org.apache.falcon.util;
+
+import java.util.Calendar;
+
+/**
+ * TimeUnit used for Date operations.
+ */
+public enum TimeUnit {
```
sandeepSamudrala: Can this class be renamed to something else to avoid ambiguity between this class and java.util.concurrent.TimeUnit ?
Can this be renamed to CalendarUnit or so?

oozie/src/main/java/org/apache/falcon/oozie/process/NativeOozieProcessWorkflowBuilder.java
```diff
@@ -0,0 +1,222 @@
...
+
+
+        DateUtil.setTimeZone(entity.getTimezone().getID());
+        ExpressionHelper.setReferenceDate(new Date(getNominalTime().getMillis()));
+        elProps.putAll(getInputProps(cluster));
+        elProps.putAll(getOutputProps());
+        elProps.putAll(evalProperties());
+        Properties buildProps  = build(cluster, buildPath);
+        buildProps.putAll(elProps);
+        copyPropsWithoutOverride(buildProps, suppliedProps);
```
sandeepSamudrala: Are suppliedProps those which are given during schedule time? If so should they not override the one in entity?

oozie/src/main/java/org/apache/falcon/oozie/process/NativeOozieProcessWorkflowBuilder.java
```diff
@@ -0,0 +1,222 @@
...
+                    if (loc.getType() != LocationType.DATA) {
+                        continue;
+                    }
+                    List<String> paths = new ArrayList<>();
+                    List<Date> instanceTimes = EntityUtil.getEntityInstanceTimes(feed, cluster.getName(),
+                            startTime, endTime); // test when startTime and endTime are equal.
+                    for (Date instanceTime : instanceTimes) {
+                        String path = EntityUtil.evaluateDependentPath(loc.getPath(), instanceTime);
+                        if (StringUtils.isNotBlank(partition)) {
+                            if (!path.endsWith("/") && !partition.startsWith("/")) {
```
sandeepSamudrala: why is this required? Even "//" will work during listing.
And also let user know they way he is giving it and not mask it to correct it.

scheduler/src/main/java/org/apache/falcon/execution/ProcessExecutionInstance.java
```diff
@@ -120,14 +125,40 @@ private void registerForNotifications(boolean isResume) throws FalconException {
...
                 List<Location> locations = FeedHelper.getLocations(cluster, feed);
                 for (Location loc : locations) {
                     if (loc.getType() != LocationType.DATA) {
                         continue;
                     }
-                    paths.add(new Path(loc.getPath()));
+                    List<Date> instanceTimes = EntityUtil.getEntityInstanceTimes(feed, cluster.getName(),
+                            startTime, endTime);
+                    for (Date instanceTime : instanceTimes) {
+                        String path = EntityUtil.evaluateDependentPath(loc.getPath(), instanceTime);
```
sandeepSamudrala: how will latest be supported in this case? or else latest is not supported with this pull req?

scheduler/src/main/java/org/apache/falcon/execution/ProcessExecutionInstance.java
```diff
@@ -120,14 +125,40 @@ private void registerForNotifications(boolean isResume) throws FalconException {
...
+                        if (feed.getAvailabilityFlag() != null && !feed.getAvailabilityFlag().isEmpty()) {
+                            if (!path.endsWith("/")) {
+                                path = path + "/";
```
sandeepSamudrala: why is this required? Even "//" will work during listing.
And also let user know they way user has given it and not mask it to correct it.

scheduler/src/main/java/org/apache/falcon/execution/SchedulerUtil.java
```diff
@@ -27,6 +28,14 @@
 
     private static final long MINUTE_IN_MS = 60 * 1000L;
     private static final long HOUR_IN_MS = 60 * MINUTE_IN_MS;
+    public static final String MINUTELY_PROCESS_FREQUENCY_POLLING_IN_MILLIS =
+            "falcon.scheduler.minutely.process.polling.frequency.millis";
```
sandeepSamudrala: This seems too expanded. having multiple polling frequencies for each frequency may not be a good idea.

I feel falcon should start considering polling frequency from user as user is more aware on the input data availability. This we can create a jira and discuss. 
But for now can we have a common polling frequency for all. We might have to later deprecate these properties if we start having polling frequency in the process.

scheduler/src/main/java/org/apache/falcon/state/InstanceState.java
```diff
@@ -270,7 +270,11 @@ public STATE nextTransition(EVENT event) throws InvalidStateTransitionException
 
     @Override
     public String toString() {
-        return instance.getId().toString() + "STATE: " + currentState.toString();
+        StringBuilder output = new StringBuilder();
+        if (instance.getId() != null) {
+            output.append(instance.getId());
+        }
+        return output.toString() + "STATE: " + currentState.toString();
```
sandeepSamudrala: output.append("STATE").append(currentState.toString()).toString?

sandeepSamudrala:
Directionally looks good. I had few comments.
Please rebase your patch.

common/src/main/java/org/apache/falcon/entity/EntityUtil.java
```diff
@@ -1082,4 +1091,15 @@ public static JOBPRIORITY getPriority(Process process) {
         }
         return JOBPRIORITY.NORMAL;
     }
+
```
ajayyadava: I understand it is a simple and straightforward method but since it is a public utility method, it will be useful to add a bit of documentation on what this method does and when to use it?

common/src/main/java/org/apache/falcon/expression/ExpressionHelper.java
```diff
@@ -257,4 +260,25 @@ public static String substitute(String originalValue, Properties properties) {
         return originalValue;
     }
 
+    public static String formatTime(String dateTimeStr, String format) throws ParseException {
```
ajayyadava: Please add documentation on what it does and when to use this method.

common/src/main/java/org/apache/falcon/util/DateUtil.java
```diff
@@ -39,8 +47,37 @@
 
     public static final long HOUR_IN_MILLIS = 60 * 60 * 1000;
 
+    private static final Pattern GMT_OFFSET_COLON_PATTERN = Pattern.compile("^GMT(\\-|\\+)(\\d{2})(\\d{2})$");
```
ajayyadava: documentation please.

common/src/main/java/org/apache/falcon/util/DateUtil.java
```diff
@@ -99,4 +136,142 @@ public static Date now() {
...
+
+    /**
+     * Parses a datetime in ISO8601 format in the Oozie processing timezone.
+     *
+     * @param s string with the datetime to parse.
+     * @return the corresponding {@link java.util.Date} instance for the parsed date.
+     * @throws java.text.ParseException thrown if the given string was
+     * not an ISO8601 value for the Oozie processing timezon.
+     */
+    public static Date parseDateOozieTZ(String s) throws ParseException {
```
ajayyadava: Why is it related to Oozie?

common/src/main/java/org/apache/falcon/util/DateUtil.java
```diff
@@ -99,4 +136,142 @@ public static Date now() {
...
         return new Date(1000L * seconds + date.getTime());
     }
+
+    /**
+     * Parses a datetime in ISO8601 format in the Oozie processing timezone.
+     *
+     * @param s string with the datetime to parse.
+     * @return the corresponding {@link java.util.Date} instance for the parsed date.
+     * @throws java.text.ParseException thrown if the given string was
+     * not an ISO8601 value for the Oozie processing timezon.
```
ajayyadava: typo: timezone.

common/src/main/java/org/apache/falcon/util/DateUtil.java
```diff
@@ -99,4 +136,142 @@ public static Date now() {
...
+            throw new ParseException("Could not parse [" + s + "] using [" + activeTimeMask + "] mask",
+                    pos.getErrorIndex());
+        }
+        if (s.length() > pos.getIndex()) {
+            throw new ParseException("Correct datetime string is followed by invalid characters: " + s, pos.getIndex());
+        }
+        return d;
+    }
+
+    private static DateFormat getISO8601DateFormat(TimeZone tz, String mask) {
```
ajayyadava: Can we document what is mask here?

common/src/main/java/org/apache/falcon/util/DateUtil.java
```diff
@@ -39,8 +47,37 @@
...
+
+    private static final Pattern VALID_TIMEZONE_PATTERN = Pattern.compile("^UTC$|^GMT(\\+|\\-)\\d{4}$");
+
+    private static final String ISO8601_TZ_MASK_WITHOUT_OFFSET = "yyyy-MM-dd'T'HH:mm";
+    private static boolean entityInUTC = true;
+
     private DateUtil() {}
 
+    /**
+     * Configures the Datetime parsing with Oozie processing timezone.
```
ajayyadava: Why is it related to Oozie?

common/src/main/java/org/apache/falcon/util/DateUtil.java
```diff
@@ -39,8 +47,37 @@
...
+     *
+     */
+    public static void setTimeZone(String tz) throws FalconException {
```
ajayyadava: It seems you are storing a state in a utility class and it doesn't seem thread safe either.

common/src/main/java/org/apache/falcon/util/DateUtil.java
```diff
@@ -99,4 +136,142 @@ public static Date now() {
...
+    }
+
+    /**
+     * Formats a {@link java.util.Calendar} as a string in ISO8601 format using Oozie processing timezone.
+     *
+     * @param c {@link java.util.Calendar} to format.
+     * @return the ISO8601 string for the given date, <code>NULL</code> if the {@link java.util.Calendar} instance was
+     * <code>NULL</code>
+     */
+    public static String formatDateOozieTZ(Calendar c) {
```
ajayyadava: Oozie?

common/src/main/java/org/apache/falcon/util/DateUtil.java
```diff
@@ -99,4 +136,142 @@ public static Date now() {
...
+    public static Calendar getCalendar(String dateString, TimeZone tz) throws Exception {
+        Date date = DateUtil.parseDateOozieTZ(dateString);
+        Calendar calDate = Calendar.getInstance();
+        calDate.setTime(date);
+        calDate.setTimeZone(tz);
+        return calDate;
+    }
+
+    /**
+     * Formats a {@link java.util.Calendar} as a string in ISO8601 format using Oozie processing timezone.
```
ajayyadava: Oozie?

oozie/src/main/java/org/apache/falcon/oozie/process/NativeOozieProcessWorkflowBuilder.java
```diff
@@ -0,0 +1,222 @@
...
+
+    private static final ExpressionHelper EXPRESSION_HELPER = ExpressionHelper.get();
+    private static final String INSTANCE_FORMAT = "yyyy-MM-dd-HH-mm";
+
+    public NativeOozieProcessWorkflowBuilder(org.apache.falcon.entity.v0.process.Process entity) {
+        super(entity);
+    }
+
+    @Override
+    public java.util.Properties build(org.apache.falcon.entity.v0.cluster.Cluster cluster,
```
ajayyadava: fully qualified class name should not be required.

common/src/main/java/org/apache/falcon/entity/EntityUtil.java
pavankumar526: Ok will add.

common/src/main/java/org/apache/falcon/expression/ExpressionHelper.java
pavankumar526: Ok

common/src/main/java/org/apache/falcon/util/DateUtil.java
pavankumar526: will remove

common/src/main/java/org/apache/falcon/util/DateUtil.java
pavankumar526: As we have to support EL expression supported by oozie, i have copied from oozie code base and added here

common/src/main/java/org/apache/falcon/util/DateUtil.java
pavankumar526: will fix

common/src/main/java/org/apache/falcon/util/DateUtil.java
pavankumar526: will remove

common/src/main/java/org/apache/falcon/util/DateUtil.java
pavankumar526: All this validations used to happen in oozie while resolving expressions. We just added in falcon now.

common/src/main/java/org/apache/falcon/util/DateUtil.java
pavankumar526: will remove

oozie/src/main/java/org/apache/falcon/oozie/process/NativeOozieProcessWorkflowBuilder.java
pavankumar526: will rename

oozie/src/main/java/org/apache/falcon/oozie/process/NativeOozieProcessWorkflowBuilder.java
pavankumar526: In my opinion they should not override the build Props and other cluster level props

pavankumar526: I have taken this from existing one which is happening in oozie schedule , don't want to break backward compatibility.

oozie/src/main/java/org/apache/falcon/oozie/process/NativeOozieProcessWorkflowBuilder.java
pavankumar526: We are just adding partition to feed instance path. While adding partition, it should be prepend with "/" right ?

scheduler/src/main/java/org/apache/falcon/execution/ProcessExecutionInstance.java
pavankumar526: It will come as new patch. I have worked on that will submit after this

scheduler/src/main/java/org/apache/falcon/execution/ProcessExecutionInstance.java
pavankumar526: Data Availability service need exact paths, it won't do list status. If we check whether path is available or not we should pass FeedPath + _SUCCEESS to it.

scheduler/src/main/java/org/apache/falcon/execution/SchedulerUtil.java
pavankumar526: This is just for initial go, later we can discuss and see what all we can add and discuss how it will go.

scheduler/src/main/java/org/apache/falcon/state/InstanceState.java
pavankumar526: My bad will change

oozie/src/main/java/org/apache/falcon/oozie/process/NativeOozieProcessWorkflowBuilder.java
sandeepSamudrala: agreed.

sandeepSamudrala:
:+1:   LGTM
org.apache.falcon.notification.service.SchedulerServiceTest seems to be failing in the jenkin merge.

pavankumar526:
The test failure is not related to this. It was failing some time back also

PR #58, opened 2016-03-01
FALCON-1826 Execution order not honoured when instances are KILLED

pallavi-rao:
The problem was that the unregister for schedule did not unregister the instance immediately. It would add to instancesToIgnore and would get ignored while scheduling. This causes issues as instancesToIgnore will need to be scanned at more than one place. 

The fix is to remove the instance from "awaitingSchedule" list synchronously for unregister.

### DIFF
--- a/scheduler/src/main/java/org/apache/falcon/notification/service/impl/SchedulerService.java
+++ b/scheduler/src/main/java/org/apache/falcon/notification/service/impl/SchedulerService.java
@@ -17,7 +17,6 @@
  */
 package org.apache.falcon.notification.service.impl;
 
-import com.google.common.cache.Cache;
 import com.google.common.cache.CacheBuilder;
 import com.google.common.cache.CacheLoader;
 import com.google.common.cache.LoadingCache;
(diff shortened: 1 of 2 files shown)


scheduler/src/main/java/org/apache/falcon/notification/service/impl/SchedulerService.java
```diff
@@ -96,22 +94,26 @@ public void register(NotificationRequest notifRequest) throws NotificationServic
...
+    public void unregister(NotificationHandler handler, ID listenerID) throws NotificationServiceException {
         // If ID is that of an entity, do nothing
         if (listenerID instanceof InstanceID) {
-            // Not efficient to iterate over elements to remove this. Add to ignore list.
-            instancesToIgnore.put((InstanceID) listenerID, new Object());
+            try {
+                InstanceID instanceID = (InstanceID) listenerID;
+                SortedMap<Integer, ExecutionInstance> instances = executorAwaitedInstances.get(instanceID
+                            .getEntityClusterID());
+                synchronized (instances) {
```
pavankumar526: Instances can be null or empty in case of no instances in ready. Null check and empty check is required here right ?

pallavi-rao: Yep. Will add the check.

pavankumar526:
Looks good +1  will merge this if there no further comments

PR #59, opened 2016-03-02
FALCON-1825 Process end time not exclusive in case of Native Scheduler

pavankumar526:


scheduler/src/main/java/org/apache/falcon/notification/service/impl/AlarmService.java
```diff
@@ -87,7 +87,7 @@ public void register(NotificationRequest notificationRequest) throws Notificatio
         DateTime nextStartTime = request.getStartTime();
         DateTime endTime;
         if (request.getEndTime().isBefore(currentTime)) {
-            endTime = request.getEndTime();
+            endTime = request.getEndTime().minusMinutes(1);
```
pallavi-rao: Should we minus the minute after if-else? Also, this change is causing some UT failures, please fix those.

pavankumar526: No we shouldn't, We can run current time instance if time is current.

pavankumar526: I will fix test case and update patch.

scheduler/src/main/java/org/apache/falcon/notification/service/impl/AlarmService.java
```diff
@@ -109,7 +109,7 @@ public void register(NotificationRequest notificationRequest) throws Notificatio
             }
         }
         // All past events have been scheduled. Nothing to schedule in the future.
-        if (request.getEndTime().isBefore(nextStartTime)) {
+        if (endTime.isBefore(nextStartTime)) {
```
pallavi-rao: request.getEndTime is also being used in creating trigger for the quartz job. Doesn't that need to be changed or is it already exclusive in quartz?

pavankumar526: It is exclusive in quartz, that why not added.

pallavi-rao: ok. +1 for the patch then.

pavankumar526: Will merge shortly

PR #61, opened 2016-03-03
FALCON-1836 - Import from database to HCatalog

vramachan:


ajayyadava:
@vramachan This is very useful and thanks for doing it. While I am migrating the existing code to lifecycle, I understand that we have to cater to user requests for new features also along with migration. However, this keeps adding more work on migration as well. I think it will help in the transition if we decide on a no more feature addition to existing import / export along with a timeline for completing the transition so that the new features additions don't have to wait too long either. Makes sense?

vramachan:
@ajayyadava Agree with you. We need to address few crucial items (like this one that fixes working with HCatalog) before the next release. I think by next week, we can safely pause on adding new features until we have it converted to lifecycle. Lets sync up on the timelines for these on Tuesday.

oozie/src/main/java/org/apache/falcon/oozie/ImportExportCommon.java
```diff
@@ -70,4 +83,37 @@ static StringBuilder buildUserPasswordArg(StringBuilder builder, StringBuilder s
...
+
+    public static void addHCatalogProperties(Properties props, Feed entity, Cluster cluster,
+        Path buildPath, WORKFLOWAPP workflow, OozieOrchestrationWorkflowBuilder<Feed> wBuilder)
+        throws FalconException {
+        if (FeedHelper.getStorageType(entity, cluster) == Storage.TYPE.TABLE) {
+            addHCatalogShareLibs(props);
+            addMetastoreURI(props, cluster);
+            if (SecurityUtil.isSecurityEnabled()) {
+                // add hcatalog credentials for secure mode and add a reference to each action
+                wBuilder.addHCatalogCredentials(workflow, cluster,
```
bvellanki: Minor nit : This is not a problem in your code. addHCatalogCredentials method uses HiveUtil.java which has a public var METASTROE_URI. If you are submitting another patch, can you  please correct this variable. If not, we can fix it later.

webapp/src/test/java/org/apache/falcon/resource/AbstractTestContext.java
```diff
@@ -33,6 +33,8 @@
     public static final String FEED_TEMPLATE1 = "/feed-template1.xml";
     public static final String FEED_TEMPLATE2 = "/feed-template2.xml";
     public static final String FEED_TEMPLATE3 = "/feed-template3.xml";
+    public static final String FEED_TEMPLATE4 = "/feed-template4.xml";
```
bvellanki: FEED_TEMPLATE4 was added but never used.

oozie/src/main/java/org/apache/falcon/oozie/DatabaseImportWorkflowBuilder.java
```diff
@@ -22,21 +22,34 @@
...
+import com.google.common.base.Splitter;
+import org.apache.hadoop.fs.Path;
+
 /**
  * Builds Datasource import workflow for Oozie.
  */
 
 public class DatabaseImportWorkflowBuilder extends ImportWorkflowBuilder {
+
+    private static final Set<String> FALCON_IMPORT_SQOOP_ACTIONS = new HashSet<String>(
```
bvellanki: FALCON_IMPORT_SQOOP_ACTIONS is never used

oozie/src/main/java/org/apache/falcon/oozie/ImportExportCommon.java
```diff
@@ -70,4 +83,37 @@ static StringBuilder buildUserPasswordArg(StringBuilder builder, StringBuilder s
         }
         return builder;
     }
+
+    public static void addHCatalogProperties(Properties props, Feed entity, Cluster cluster,
+        Path buildPath, WORKFLOWAPP workflow, OozieOrchestrationWorkflowBuilder<Feed> wBuilder)
```
bvellanki: Where is buildPath being used?

bvellanki:
There are checkstyle failures when I apply and build the patch. 
[INFO] Starting audit...
/Users/bvellanki/SourceCode/apache-falcon/falcon/webapp/src/test/java/org/apache/falcon/lifecycle/FeedImportIT.java:24:8: Unused import - org.apache.falcon.catalog.HiveCatalogService.
/Users/bvellanki/SourceCode/apache-falcon/falcon/webapp/src/test/java/org/apache/falcon/lifecycle/FeedImportIT.java:27:8: Unused import - org.apache.falcon.entity.v0.feed.CatalogTable.
Audit done.

oozie/src/main/java/org/apache/falcon/oozie/ImportExportCommon.java
vramachan: Renamed.

oozie/src/main/java/org/apache/falcon/oozie/DatabaseExportWorkflowBuilder.java
```diff
@@ -169,4 +182,50 @@ private StringBuilder buildNumMappers(StringBuilder builder, Map<String, String>
...
         return FeedHelper.getExportArguments(feedCluster);
     }
+
+    private StringBuilder buildExportTableArg(StringBuilder builder, CatalogTable catalog) throws FalconException {
+
+        LOG.info("Catalog URI {}", catalog.getUri());
+        builder.append("--skip-dist-cache").append(ImportExportCommon.ARG_SEPARATOR);
+        Iterator<String> itr = Splitter.on("#").split(catalog.getUri()).iterator();
+        String dbTable = itr.next();
+        String partitions = itr.next();
```
bvellanki: Can there be no partitions? If yes itr.next() will return null and getPartitionKeyValues method will throw exception. If partitions are always provided, "if" condition in line 207 seems unnecessary

bvellanki:
@vramachan  : Please fix the possible nullpointer exception. Rest of the code looks good. I will commit once you make this fix and once we get another +1 from community

oozie/src/main/java/org/apache/falcon/oozie/DatabaseExportWorkflowBuilder.java
vramachan: @bvellanki the partitions are must; validated by submit

ERROR: Bad Request;default/java.net.URISyntaxException: Partition details are missing: catalog:default:hcat_simple_1
CausedBy: Partition details are missing: catalog:default:hcat_simple_1

bvellanki:
@vramachan If partitions are must and feed is validated during submit, I give +1 for this patch.

peeyushb:
@vramachan Patch looks good. +1 .

vramachan:
@ajayyadava  In response to your prior comment about migration to lifecycle: This import/export patch is committed and there are no further improvements planned for the next 2-3 weeks. Let me know if this is good enough for the Lifecycle migration. Should any urgent fix/improvement comes up, I will coordinate with you.

PR #62, opened 2016-03-04
FALCON-1845 Retries fix when entity was deleted during instance rerun

pavankumar526:


rerun/src/main/java/org/apache/falcon/rerun/handler/RetryHandler.java
```diff
@@ -112,6 +112,8 @@ public void onSuccess(WorkflowExecutionContext context) throws FalconException {
     public void onFailure(WorkflowExecutionContext context) throws FalconException {
         // Re-run does not make sense when killed by user.
         if (context.isWorkflowKilledManually()) {
+            LOG.debug("User workflow: {} killed manually of instance: {} belongs to entity: {}, so retry won't happen ",
```
pallavi-rao: Minor nit on the phrasing of the message. Can we rephrase to "Workflow {}; Instance {}; Entity {}, killed manually by user. Will not retry."

pallavi-rao:
+1 after from the minor nit.

rerun/src/main/java/org/apache/falcon/rerun/handler/RetryHandler.java
```diff
@@ -112,6 +112,8 @@ public void onSuccess(WorkflowExecutionContext context) throws FalconException {
     public void onFailure(WorkflowExecutionContext context) throws FalconException {
         // Re-run does not make sense when killed by user.
         if (context.isWorkflowKilledManually()) {
+            LOG.debug("Workflow: {} Instance: {} Entity: {}, killed manually by user. Will not retry.",
```
pavankumar526: Changed based on comments

rerun/src/main/java/org/apache/falcon/rerun/handler/AbstractRerunConsumer.java
```diff
@@ -88,7 +88,6 @@ public void run() {
                     LOG.warn("Entity {} of type {} doesn't exist in config store. Rerun "
                                     + "cannot be done for workflow ", message.getEntityName(),
                             message.getEntityType(), message.getWfId());
-                    return;
```
sandeepSamudrala: It should be continue else the message below would get logged. (Error in rerun consumer) ?

sandeepSamudrala:
:+1:  other than that one comment

rerun/src/main/java/org/apache/falcon/rerun/handler/AbstractRerunConsumer.java
pavankumar526: Makes sense will change

PR #63, opened 2016-03-04
FALCON-1846 Fixing ProcessInstanceStatusTest,ProcessInstanceResumeTest,EntityDryRunTest

pragya-mittal:
FALCON-1846 Fixing ProcessInstanceStatusTest,ProcessInstanceResumeTest,EntityDryRunTest


falcon-regression/merlin-core/src/main/java/org/apache/falcon/regression/core/util/InstanceUtil.java
```diff
@@ -516,7 +516,7 @@ public static String getOutputFolderForInstanceForReplication(OozieClient oozieC
     private static String getReplicatedFolderFromInstanceRunConf(String runConf) {
         String inputPathExample = getReplicationFolderFromInstanceRunConf(runConf).get(0);
         String postFix = inputPathExample.substring(inputPathExample.length() - 7, inputPathExample.length());
-        return getReplicatedFolderBaseFromInstanceRunConf(runConf) + postFix;
+        return getReplicatedFolderBaseFromInstanceRunConf(runConf) + "/" + postFix;
     }
```
muraliramasami: @pragya-mittal  Can we use the OSUtil.concat() here?

pragya-mittal: Cannot use this function. It replaces all the occurrences of '//' by '/'. My input contains hdfs://randomPath which is converted to hdfs:/randomPath which is not the expected value.

pragya-mittal: Removing ProcessInstanceStatusTest changes as discussed.

muraliramasami: Thanks @pragya-mittal

muraliramasami:
Thanks @pragya-mittal  . +1 Patch Looks good to me.

PR #64, opened 2016-03-07
FALCON-1847 Execution order not honored when instances are suspended/resumed

pallavi-rao:
Have made the following changes:

1. Order of instances on which we perform action is sorted in desc only for status and params. Else, older instances should be suspended/resumed first.
2. Status is now read from DB and only additional info is obtained from Oozie.
3. Listening to suspend notification too, so further instances can be scheduled.
4. Resume does not immediately resume the instance in order to honor concurrency.

### DIFF
--- a/scheduler/src/test/java/org/apache/falcon/notification/service/SchedulerServiceTest.java
+++ b/scheduler/src/test/java/org/apache/falcon/notification/service/SchedulerServiceTest.java
@@ -45,6 +45,7 @@ import org.apache.falcon.util.StartupProperties;
 import org.apache.falcon.util.StateStoreProperties;
 import org.apache.falcon.workflow.engine.DAGEngine;
 import org.apache.falcon.workflow.engine.DAGEngineFactory;
+import org.apache.falcon.workflow.engine.FalconWorkflowEngine;
 import org.apache.oozie.client.WorkflowJob;
 import org.joda.time.DateTime;
 import org.mockito.Mockito;
@@ -56,6 +57,7 @@ import org.testng.annotations.Test;
 
 import java.util.ArrayList;
 import java.util.Date;
+import java.util.Properties;
 
 import static org.apache.falcon.state.InstanceState.STATE;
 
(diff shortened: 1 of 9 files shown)


scheduler/src/test/java/org/apache/falcon/notification/service/SchedulerServiceTest.java
```diff
@@ -305,6 +307,25 @@ public PRIORITY getPriority() {
...
+        Properties resumeProps = new Properties();
+        resumeProps.setProperty(FalconWorkflowEngine.FALCON_RESUME, "true");
+        instance1.setProperties(resumeProps);
+        instance1.setExternalID("123");
+        SchedulerService.JobScheduleRequestBuilder request = (SchedulerService.JobScheduleRequestBuilder)
+                scheduler.createRequestBuilder(handler, instance1.getId());
+        request.setInstance(instance1);
+        scheduler.register(request.build());
+        Thread.sleep(100);
+        Assert.assertEquals(((MockDAGEngine)mockDagEngine).getTotalResumes(instance1), new Integer(1));
```
sandeepSamudrala: Can there be an assertion for running instances too?

sandeepSamudrala: getTotalResumes is only giving resumeInvocations, but not necessarily that these instances were actually resumed,
Should not the assertion be on the status to be !suspend?

pallavi-rao: This is the unit test for SchedulerService which is not aware of statuses of instances. For the status check I have updated InstanceSchedulerManagerJerseyIT accordingly.

sandeepSamudrala: I had missed it. makes sense. Thanks @pallavi-rao  for the clarification.

sandeepSamudrala:
:+1:

pallavi-rao:
Thanks @sandeepSamudrala for the review. Will merge if no further comments.

PR #71, opened 2016-03-15
FALCON-1838 : Add Export instances to graphdb for Lineage tracking

vramachan:
@falcon Please review


oozie/src/main/java/org/apache/falcon/oozie/DatabaseExportWorkflowBuilder.java
```diff
@@ -169,4 +187,50 @@ private StringBuilder buildNumMappers(StringBuilder builder, Map<String, String>
...
+            StringBuilder partitionValues = new StringBuilder();
+            for (Map.Entry<String, String> e : partitionsMap.entrySet()) {
+                partitionKeys.append(e.getKey());
+                partitionKeys.append(',');
+                partitionValues.append(String.format("${coord:dataInPartitionMin('%s','%s')}",
+                        FeedExportCoordinatorBuilder.EXPORT_DATAIN_NAME,
+                        e.getKey()));
+                partitionValues.append(',');
+            }
+            if (partitionsMap.size() > 0) {
```
deepakbarr: This condition is already checked in outer if condition.

PR #73, opened 2016-03-16
FALCON-1852 Make optional input to a process truly optional

pallavi-rao:
The following changes have been made:
1. Creation of "empty dir" under the staging path of a cluster (during cluster creation).
2. Modify OozieELExtensions to look for availability flag and use only those instances that have the entire dataset. Use "empty dir" when no instances resolve.
3. Updated UT and IT for additional validation.

Tested manually with and without availabilityFlag supplied.

### DIFF
--- a/oozie-el-extensions/src/main/java/org/apache/oozie/extensions/OozieELExtensions.java
+++ b/oozie-el-extensions/src/main/java/org/apache/oozie/extensions/OozieELExtensions.java
@@ -20,11 +20,14 @@ package org.apache.oozie.extensions;
 
 import org.apache.commons.lang3.StringUtils;
 import org.apache.hadoop.conf.Configuration;
+import org.apache.oozie.client.OozieClient;
 import org.apache.oozie.command.coord.CoordCommandUtils;
 import org.apache.oozie.coord.CoordELEvaluator;
 import org.apache.oozie.coord.CoordELFunctions;
 import org.apache.oozie.coord.SyncCoordAction;
 import org.apache.oozie.coord.SyncCoordDataset;
+import org.apache.oozie.dependency.ActionDependency;
+import org.apache.oozie.dependency.DependencyChecker;
 import org.apache.oozie.util.ELEvaluator;
 import org.apache.oozie.util.ParamChecker;
 import org.apache.oozie.util.XLog;
@@ -34,8 +37,10 @@ import org.jdom.Text;
 
 import java.text.DateFormat;
 import java.text.SimpleDateFormat;
+import java.util.ArrayList;
 import java.util.Calendar;
 import java.util.Date;
+import java.util.List;
 import java.util.TimeZone;
 
 /**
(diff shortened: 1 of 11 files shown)


oozie-el-extensions/src/main/java/org/apache/oozie/extensions/OozieELExtensions.java
```diff
@@ -72,15 +77,31 @@ public static String ph3_dataIn(String dataInName, String part) {
...
+            Configuration conf = new Configuration();
+            conf.set(OozieClient.USER_NAME, (String)eval.getVariable(OozieClient.USER_NAME));
             try {
                 ELEvaluator instEval = CoordELEvaluator.createInstancesELEvaluator(dsEle, appInst, conf);
                 StringBuilder instances = new StringBuilder();
-                CoordCommandUtils.resolveInstanceRange(dsEle, instances , appInst, conf, instEval);
-                uristr = CoordCommandUtils.createEarlyURIs(dsEle, instances.toString(),
-                        new StringBuilder(), new StringBuilder());
-                uristr = uristr.replace(CoordELFunctions.INSTANCE_SEPARATOR, ",");
+                StringBuilder urisWithDoneFlag = new StringBuilder();
```
sriksun: There are no test cases for this. Can we please add a few. This is the critical piece of the patch

webapp/src/test/java/org/apache/falcon/resource/EntityManagerJerseyIT.java
```diff
@@ -198,6 +198,8 @@ public void testOptionalInput() throws Exception {
         File tmpFile = TestContext.getTempFile();
         EntityType.PROCESS.getMarshaller().marshal(process, tmpFile);
         schedule(context);
+        waitForStatus(EntityType.PROCESS.name(), context.getProcessName(), START_INSTANCE,
```
sriksun: Docs update would be very welcome

oozie-el-extensions/src/main/java/org/apache/oozie/extensions/OozieELExtensions.java
```diff
@@ -72,15 +77,31 @@ public static String ph3_dataIn(String dataInName, String part) {
...
+                CoordCommandUtils.resolveInstanceRange(dsEle, instances, appInst, conf, instEval);
+                CoordCommandUtils.createEarlyURIs(dsEle, instances.toString(),
+                        new StringBuilder(), urisWithDoneFlag);
+                XLog.getLog(OozieELExtensions.class).debug("Resolved instances for " + dataInName + " : "
+                        + urisWithDoneFlag.toString());
+                // Check if availability flags are present for each instance.
```
sandeepSamudrala: This may not be the case for all feeds. A feed may not have a availability Flag. In which case,  mere presence of the directory is good enough.

oozie-el-extensions/src/main/java/org/apache/oozie/extensions/OozieELExtensions.java
```diff
@@ -72,15 +77,31 @@ public static String ph3_dataIn(String dataInName, String part) {
...
+                ActionDependency actionDep = DependencyChecker.checkForAvailability(urisWithDoneFlag.toString(),
```
sandeepSamudrala: Ignore. I had just checked the oozie code, urisWithDoneFlag would be populated only for those feeds with availability flag. Correct me if I am wrong, else ignore my previous comment

pallavi-rao: That is right @sandeepSamudrala, if the availabilityFlag is empty, the method checks if the dir exists. So, we are good there too.

webapp/src/test/java/org/apache/falcon/resource/EntityManagerJerseyIT.java
pallavi-rao: One case where docs are ahead of code. The Entity Specification already says this -
"Optional Inputs

User can mention one or more inputs as optional inputs. In such cases the job does not wait on those inputs which are mentioned as optional. If they are present it considers them otherwise continue with the compulsory ones. " - http://falcon.apache.org/EntitySpecification.html#Process_Specification

Let me know if I should add more.

oozie-el-extensions/src/main/java/org/apache/oozie/extensions/OozieELExtensions.java
pallavi-rao: I tried to add a test case (not hard enough though), DependencyChecker.checkForAvailability is a static method and can't be mocked easily. Will see if I can work around that. Else, will try and use powermock.

oozie-el-extensions/src/main/java/org/apache/oozie/extensions/OozieELExtensions.java
```diff
@@ -72,15 +77,31 @@ public static String ph3_dataIn(String dataInName, String part) {
...
+                        conf, false);
+                String doneFlag = (String) eval.getVariable(dataInName + ".done-flag");
+                uristr = StringUtils.join(stripDoneFlag(actionDep.getAvailableDependencies(), doneFlag), ",");
+                // If no instances are present, point the optional input to empty dir.
+                if (StringUtils.isEmpty(uristr)) {
+                    String emptyDir = (String) eval.getVariable(dataInName + ".empty-dir");
```
pavankumar526: Is this empty directory  different for different feeds ? Or we are pointing common empty directory which was created in Staging directory ?

pallavi-rao: It is per cluster. So, common for all feeds in a cluster.

pavankumar526: Ok Thanks

oozie-el-extensions/src/main/java/org/apache/oozie/extensions/OozieELExtensions.java
pallavi-rao: Added test cases.

pallavi-rao:
Build failure not related to the patch:
`[INFO] Apache Falcon Distro .............................. FAILURE [  2.123 s]
[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
[INFO] Total time: 29:16 min
[INFO] Finished at: 2016-03-17T17:56:44+00:00
[INFO] Final Memory: 797M/2374M
[INFO] ------------------------------------------------------------------------
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-assembly-plugin:2.2.1:single (dist-assembly) on project falcon-distro: Failed to create assembly: Error creating assembly archive bin: Failed to retrieve numeric file attributes using: '/bin/sh -c ls -1nlaR /home/jenkins/jenkins-slave/workspace/Apache-Falcon-Pull-Request-Build/distro/../.' -> [Help 1]`

webapp/src/test/java/org/apache/falcon/resource/EntityManagerJerseyIT.java
ajayyadava: I think it will be useful to document the behavior if some instances in the input range for an optional input are available explicitly.

sandeepSamudrala:
:+1:  Patch looks good.
As already mentioned by @ajayyadava , can you add a line on optional that, only fully available data would be considered for input.

pallavi-rao:
Thanks @sandeepSamudrala , @sriksun, @pavankumar526, @ajayyadava, thanks for the review. Will merge if no further comments.

PR #74, opened 2016-03-16
FALCON-1838:Export instances are not added graph db for lineage tracking

vramachan:


common/src/main/java/org/apache/falcon/metadata/MetadataMappingService.java
```diff
@@ -345,6 +345,9 @@ private void updateInstanceStatus(final WorkflowExecutionContext context) throws
         case IMPORT:
             updateImportedFeedInstance(context);
             break;
+        case EXPORT:
+            onFeedInstanceExported(context);
```
bvellanki: Minor nit : Can we have uniform naming for similar methods?  This one should be updateExportedFeedInstance...

common/src/main/java/org/apache/falcon/metadata/InstanceRelationshipGraphBuilder.java
```diff
@@ -256,34 +256,47 @@ public void addEvictedInstance(WorkflowExecutionContext context) throws FalconEx
...
-        String feedName = context.getOutputFeedNames();
-        String feedInstanceDataPath = context.getOutputFeedInstancePaths();
+    public void addExportedInstance(WorkflowExecutionContext context) throws FalconException {
+        addImportExportInstanceHelper(context, RelationshipLabel.DATASOURCE_EXPORT_EDGE);
+    }
+
+    private void addImportExportInstanceHelper(WorkflowExecutionContext context,
+                                               RelationshipLabel label) throws FalconException {
+        String feedName = (label == RelationshipLabel.DATASOURCE_IMPORT_EDGE)
+                ? context.getOutputFeedNames() : context.getInputFeedNames();
```
bvellanki: Shouldnt this be 

String feedName = (label == RelationshipLabel.DATASOURCE_IMPORT_EDGE)
                ? context.getInputFeedNames() : context.getOutputFeedNames();

common/src/main/java/org/apache/falcon/metadata/InstanceRelationshipGraphBuilder.java
```diff
@@ -256,34 +256,47 @@ public void addEvictedInstance(WorkflowExecutionContext context) throws FalconEx
...
+        String feedInstanceDataPath = (label == RelationshipLabel.DATASOURCE_IMPORT_EDGE)
```
bvellanki: Same as above

common/src/main/java/org/apache/falcon/metadata/MetadataMappingService.java
vramachan: Looks like the IMPORT and others got refactored between my patch attempts.

common/src/main/java/org/apache/falcon/metadata/InstanceRelationshipGraphBuilder.java
vramachan: IMPORT -> gets data in to HDFS and so, it needs output path or partitions generated by Coordinator to write to and context.getOutputFeedNames()

common/src/main/java/org/apache/falcon/metadata/InstanceRelationshipGraphBuilder.java
vramachan: EXPORT -> gets data out of Hadoop -> so, need the input path or HCAT partitions to consume and so context.getInputFeedNames()

common/src/main/java/org/apache/falcon/metadata/InstanceRelationshipGraphBuilder.java
bvellanki: Ah, I got confused due to lack of context. When I read the full file, it makes complete sense. Thank you

bvellanki:
+1 @vramachan patch looks good.

common/src/main/java/org/apache/falcon/metadata/InstanceRelationshipGraphBuilder.java
```diff
@@ -256,34 +256,47 @@ public void addEvictedInstance(WorkflowExecutionContext context) throws FalconEx
...
         String feedInstanceName = getFeedInstanceName(feedName, sourceClusterName,
                 feedInstanceDataPath, context.getNominalTimeAsISO8601());
-        Vertex feedInstanceVertex = addFeedInstance(
-                feedInstanceName, context, feedName, context.getSrcClusterName());
-
+        Vertex feedInstanceVertex = findVertex(feedInstanceName, RelationshipType.FEED_INSTANCE);
+
+        LOG.info("Vertex exists? name={}, type={}, v={}",
+                feedInstanceName, RelationshipType.FEED_INSTANCE, feedInstanceVertex);
+        if (feedInstanceVertex == null) { // No record of instances NOT generated by Falcon
```
yzheng-hortonworks: You don't need to call findVertex to check whether the vertex exists or not. This is already handled in addVertex in addFeedInstance.

common/src/main/java/org/apache/falcon/metadata/InstanceRelationshipGraphBuilder.java
```diff
@@ -256,34 +256,47 @@ public void addEvictedInstance(WorkflowExecutionContext context) throws FalconEx
...
+    }
+
+    private void addImportExportInstanceHelper(WorkflowExecutionContext context,
+                                               RelationshipLabel label) throws FalconException {
+        String feedName = (label == RelationshipLabel.DATASOURCE_IMPORT_EDGE)
+                ? context.getOutputFeedNames() : context.getInputFeedNames();
+        String feedInstanceDataPath = (label == RelationshipLabel.DATASOURCE_IMPORT_EDGE)
+                ? context.getOutputFeedInstancePaths() : context.getInputFeedInstancePaths();
         String datasourceName = context.getDatasourceName();
         String sourceClusterName = context.getSrcClusterName();
```
yzheng-hortonworks: Should it be different value for export, i.e. context.getClusterName()? Same question for datasourceName in the previous line.

common/src/main/java/org/apache/falcon/metadata/MetadataMappingService.java
```diff
@@ -370,4 +373,8 @@ private void updateImportedFeedInstance(WorkflowExecutionContext context) throws
         LOG.info("Updating imported feed instance: {}", context.getNominalTimeAsISO8601());
         instanceGraphBuilder.addImportedInstance(context);
     }
+    private void updateExportedFeedInstance(WorkflowExecutionContext context) throws FalconException {
+        LOG.info("Adding export feed instance: {}", context.getNominalTimeAsISO8601());
```
yzheng-hortonworks: Better to use "Updating" instead of "Adding" for consistency. Also, this will be called to either add new feed instance or update the instance status.

common/src/main/java/org/apache/falcon/metadata/InstanceRelationshipGraphBuilder.java
vramachan: Looks like I overwrote your change on this. Picked up the implementation that does not call findVertex() explicitly.

common/src/main/java/org/apache/falcon/metadata/MetadataMappingService.java
vramachan: alright

common/src/main/java/org/apache/falcon/metadata/InstanceRelationshipGraphBuilder.java
vramachan: Nope. why? The context has the datasource and cluster from which data is exported out. Also, same for the IMPORT case. 

BTW, while all the comments appear in the latest version, this one appears on the older version (on the left side) :-)

yzheng-hortonworks: @vramachan Looks good to me. Thanks for clarifying the definition of source cluster and data source in export and import.

yzheng-hortonworks:
+1!

PR #75, opened 2016-03-17
FALCON-1840 Archive older definition in case of update

PraveenAdlakha:


bvellanki:
Should we undo archive if updateInternal fails?

common/src/main/java/org/apache/falcon/entity/store/ConfigurationStore.java
```diff
@@ -250,6 +250,11 @@ private synchronized void updateInternal(EntityType type, Entity entity) throws
 
     public synchronized void update(EntityType type, Entity entity) throws FalconException {
         if (updatesInProgress.get() == entity) {
+            try {
+                archive(type,entity.getName());
+            } catch (IOException e) {
+                throw new StoreAccessException(e);
+            }
             updateInternal(type, entity);
```
pallavi-rao: Update an existing IT or add a new test, please.

PraveenAdlakha:
@bvellanki  we are doing the same thing in delete also we first archive and then delete.In worst case we will have same copy which is currently running in production.

PraveenAdlakha:
Test case of archive was missing on delete.I have added that too.

peeyushb:
@PraveenAdlakha Patch looks good. +1. Is it possible to add some documentation for this so that user should get inform that along with update of entity happen, archive of entity also occur.

pallavi-rao:
+1 for the patch @PraveenAdlakha. But, as @peeyushb mentioned, updating the doc will be a good idea.

sandeepSamudrala:
:+1:  @PraveenAdlakha : The next question that user would look for is, what does happen in subsequent updates?. Can you also add that its versioned(timestamp) archival.

PraveenAdlakha:
@sandeepSamudrala  its already version when ever we archive we add timestamp to the files.

peeyushb:
Thanks @PraveenAdlakha for adding the documentation. If no one has any further comments, I will commit this patch soon.

PR #76, opened 2016-03-21
FALCON-1864 Retry event does not get removed from delay queue

pallavi-rao:


pallavi-rao:
Build failure unrelated to patch.

rerun/src/main/java/org/apache/falcon/rerun/handler/RetryConsumer.java
```diff
@@ -43,14 +43,18 @@ public RetryConsumer(T handler) {
...
-                        + " {}", message.getWfId());
+            // Can happen when user does a manual re-run in-between retries.
+            if (jobStatus.equals("RUNNING") || jobStatus.equals("PREP")) {
+                LOG.debug("Re-enqueing message in RetryHandler for workflow with same delay as "
+                        + "job status is {} for : {}", jobStatus, message.getWfId());
                 message.setMsgInsertTime(System.currentTimeMillis());
                 handler.offerToQueue(message);
                 return;
+            } else if (jobStatus.equals("SUSPENDED") || jobStatus.equals("SUCCEEDED")) {
+                LOG.debug("Not retrying workflow {} anymore as it is in {} state. ", message.getWfId(), jobStatus);
```
sandeepSamudrala: During suspended we will might have to en-queue the message, as once the instance is resumed, it might still be required to check for the instance status and then proceed further

pallavi-rao: I thought about it too. My thought process was, if it has been suspended, then, when the user resumes it, it will either fail or succeed. If it fails, the retry will anyway kick in (as a new RetryEvent). Since we derive the retry count from the runId, we won't exceed the attempts. So, I thought there is no need to constantly monitor a SUSPENDED instance for retries.

Am I missing something?

sandeepSamudrala: Makes sense. Just thinking if there could be a corner case where the instance after resuming had gone to timed out state or so because of which retry not happening.

pallavi-rao: In that case, the timeout retry should kick in based on whether user has configured it.

pavankumar526: Makes sense , so we are rerunning Failed workflow also right, which is not happening previously.

pallavi-rao: Yep. We will rerun failed or killed or timed_out (if user has chosen) workflows.

pavankumar526: +1 for the patch

sandeepSamudrala:
:+1:

rerun/src/main/java/org/apache/falcon/rerun/handler/RetryConsumer.java
PraveenAdlakha: +1

pallavi-rao:
Thanks @sandeepSamudrala, @pavankumar526, @PraveenAdlakha for the review. Will merge shortly.

PR #78, opened 2016-03-24
FALCON-1874 Import and Export fails with HDFS as src/dest

pallavi-rao:
1. job-xml was pointing to hive-site.xml even for HDFS, cause workflow to fail.
2. datain element was repeated in export coordinator, causing workflow to fail.


vramachan:
+1 
Looks good.

PraveenAdlakha:
+1 LGTM

pallavi-rao:
Thanks @vramachan, @PraveenAdlakha for the review. Will merge shortly.

PR #80, opened 2016-03-29
FALCON-1866 bug in JDBCStateStore

PraveenAdlakha:


scheduler/src/main/java/org/apache/falcon/state/store/jdbc/JDBCStateStore.java
```diff
@@ -104,10 +104,10 @@ private EntityBean getEntityBean(EntityID id) {
         Query q = entityManager.createNamedQuery("GET_ENTITY");
         q.setParameter("id", id.getKey());
         List result = q.getResultList();
+        entityManager.close();
         if (result.isEmpty()) {
             return null;
         }
-        entityManager.close();
         return ((EntityBean)result.get(0));
```
pavankumar526: Looks good +1

ajayyadava:
:+1: The failure seems to be unrelated. Will commit tomorrow unless someone has concerns.

PR #81, opened 2016-03-30
Falcon webUI returns 413

vrangan:
(Full head - Request entity too large) error when TLS is enabled in a secure cluster with AD integration)

Fix is to duplicate the SocketConnector change to increase header and request buffer for Jetty based on startup properties

### DIFF
--- a/prism/src/main/java/org/apache/falcon/util/SecureEmbeddedServer.java
+++ b/prism/src/main/java/org/apache/falcon/util/SecureEmbeddedServer.java
@@ -48,6 +48,13 @@ public class SecureEmbeddedServer extends EmbeddedServer {
         connector.setPassword(properties.getProperty("password",
                 System.getProperty("password", "falcon-prism-passwd")));
         connector.setWantClientAuth(true);
+
+        // this is to enable large header sizes when Kerberos is enabled with AD
+        final Integer bufferSize = Integer.valueOf(StartupProperties.get().getProperty(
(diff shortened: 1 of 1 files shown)


prism/src/main/java/org/apache/falcon/util/SecureEmbeddedServer.java
```diff
@@ -48,6 +48,13 @@ protected Connector getConnector(int port) {
         connector.setPassword(properties.getProperty("password",
                 System.getProperty("password", "falcon-prism-passwd")));
         connector.setWantClientAuth(true);
+
+        // this is to enable large header sizes when Kerberos is enabled with AD
+        final Integer bufferSize = Integer.valueOf(StartupProperties.get().getProperty(
+                "falcon.jetty.request.buffer.size", "16192"));
```
ajayyadava: We should add this property to startup.properties as well with default value and a comment to describe the property. This will enable users to know that to tweak this behavior they can change value of this property.

vrangan:
Failure unrelated

prism/src/main/java/org/apache/falcon/util/SecureEmbeddedServer.java
vrangan: Good idea.   we have a bunch of these properties  - will create a separate JIRA to track that

bvellanki:
+1 will commit shortly

PR #82, opened 2016-04-01
FALCON-1880 Support TDE encryption

bvellanki:
Add --skipcrccheck to distcp options for HiveDR.

### DIFF
--- a/addons/hivedr/src/main/java/org/apache/falcon/hive/HiveDROptions.java
+++ b/addons/hivedr/src/main/java/org/apache/falcon/hive/HiveDROptions.java
@@ -120,7 +120,7 @@ public class HiveDROptions {
     }
 
     public int getMaxEvents() {
-        return Integer.valueOf(context.get(HiveDRArgs.MAX_EVENTS));
+        return Integer.parseInt(context.get(HiveDRArgs.MAX_EVENTS));
     }
 
     public boolean shouldKeepHistory() {
@@ -145,6 +145,11 @@ public class HiveDROptions {
 
     public String getExecutionStage() {
         return context.get(HiveDRArgs.EXECUTION_STAGE);
+    }
+
+    public boolean isTDEEncryptionEnabled() {
+        return StringUtils.isEmpty(context.get(HiveDRArgs.TDE_ENCRYPTION_ENABLED))
+                ? false : Boolean.valueOf(context.get(HiveDRArgs.TDE_ENCRYPTION_ENABLED));
     }
 
     public boolean shouldBlock() {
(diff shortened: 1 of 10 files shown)


addons/hivedr/src/main/java/org/apache/falcon/hive/HiveDROptions.java
```diff
@@ -147,6 +147,11 @@ public String getExecutionStage() {
         return context.get(HiveDRArgs.EXECUTION_STAGE);
     }
 
+    public boolean isTDEEncryptionEnabled() {
+        return StringUtils.isEmpty(context.get(HiveDRArgs.TDE_ENCRYPTION_ENABLED))
+                ? false : Boolean.valueOf(context.get(HiveDRArgs.TDE_ENCRYPTION_ENABLED));
```
vramachan: Boolean.parseBoolean() that returns boolean?

bvellanki: if the string has a typo like "flase" , the parse would throw an error. If the string is not defined, we want to use the default value, which is false.

addons/hivedr/src/main/java/org/apache/falcon/hive/HiveDRTool.java
```diff
@@ -271,8 +271,8 @@ private void cleanStagingDirectory() throws HiveReplicationException {
     private String sourceEvents() throws Exception {
         MetaStoreEventSourcer defaultSourcer = null;
         String inputFilename = null;
-        String lastEventsIdFile = FileUtils.DEFAULT_EVENT_STORE_PATH +File.separator+inputOptions.getJobName()+"/"
-                +inputOptions.getJobName()+".id";
+        String lastEventsIdFile = FileUtils.DEFAULT_EVENT_STORE_PATH + File.separator
+                + inputOptions.getJobName() + "/" + inputOptions.getJobName() + ".id";
```
vramachan: Looks like a format issue but no code change? You can use File.separator instead of "/"

bvellanki: OK will do.

peeyushb:
@bvellanki Patch looks good. Can you add some documentation so that user must be aware of how to copy data, if TDE encryption is enabled.

vramachan:
+1 Looks good.

ajayyadava:
Can we update the documentation also please?

bvellanki:
Peeyush : Since it is TDE, user does not have to do anything different to copy data during manual bootstrap steps.
Peeyush and Ajay : I added documentation in HiveDR.twiki on how to update recipe properties when TDE encryption is enabled.

PR #83, opened 2016-04-01
Falcon-1881 : Database export does not require fields tag in entity specification

vramachan:
Database export does not require fields tag in entity specification


pallavi-rao:
Is the intention of the change to make fields (includes and excludes) optional or is it to remove the ability to specify the include/exclude fields? If it is latter, why are we doing it?

vramachan:
@pallavi-rao For export, <fields> needs to go because it is superfluous. The list of fields that gets exported to database is determined by the target table definition. If the target table definition has 3 fields, then only 3 fields from the HDFS will be exported. 

For import, we need the fields explicitly specified by the user. 

The way this is implemented now is to do a check in the EntityParser to validate the Export does not have <fields> tag.

peeyushb:
@vramachan Thanks for adding the details. Patch looks good. +1.

bvellanki:
+1 for the patch. Will commit shortly.

PR #85, opened 2016-04-04
FALCON-1883:Falcon regression build fails with minor checkstyle issues

muraliramasami:
Fixing the checkstyle issue observed during building the regression trunk


pragya-mittal:
+1

muraliramasami:
Thanks @pragya-mittal  for the review

PR #87, opened 2016-04-05
FALCON-1882 Instance status api not working via prism

PraveenAdlakha:


PraveenAdlakha:
Testing :

$ bin/falcon instance -type feed -name feed1 -status -lifecycle EVICTION,REPLICATION
Consolidated Status: SUCCEEDED

Instances:
Instance		Cluster		SourceCluster		Status		Start		End		Details					Log
-----------------------------------------------------------------------------------------------

Additional Information:
Response: local/STATUS
Request Id: local/577728771@qtp-1744005378-0 - d2355720-a9a8-45d7-9


$ bin/falcon instance -type feed -name feed1 -status -lifecycle EVICTION
Consolidated Status: SUCCEEDED

Instances:
Instance		Cluster		SourceCluster		Status		Start		End		Details					Log
-----------------------------------------------------------------------------------------------

Additional Information:
Response: local/STATUS
Request Id: local/577728771@qtp-1744005378-0 - 6a620f94-34da-47bc-a92c-f50454986da0


$ bin/falcon instance -type feed -name feed1 -status 
Consolidated Status: SUCCEEDED

Instances:
Instance		Cluster		SourceCluster		Status		Start		End		Details					Log
-----------------------------------------------------------------------------------------------

Additional Information:
Response: local/STATUS
Request Id: local/577728771@qtp-1744005378-0 - 295de95d-ba32-42b0-a71d-a7c0bbc90727

prism/src/main/java/org/apache/falcon/resource/channel/HTTPChannel.java
```diff
@@ -170,8 +171,20 @@ private String pathValue(Method method, Object... args) throws FalconException {
                     String annotationClass = paramAnnotation.annotationType().getName();
 
                     if (annotationClass.equals(QueryParam.class.getName())) {
-                        queryString.append(getAnnotationValue(paramAnnotation, "value")).
-                                append('=').append(arg).append("&");
+                        if (args[index] instanceof List){
+                            List lifecycle = (List) args[index];
```
pavankumar526: It should be valid for all list query params. why did you specify lifecycle here ?

PraveenAdlakha: I believe we cannot handle it generically here because the nature of the list that we can get in jax rs can be very different and might not implement a to string method. We might end up getting a run time exception.

Also just to point out we can refactor it right now code is not generic enough we have checks like args[index] instanceof String || args[index] instanceof Boolean || args[index] instanceof Integer
                    || args[index] instanceof List 
tomorrow a set might also come.

I am open for suggestions if you have a better way to implement it.

ajayyadava: I agree with Pavan and think this piece of code is ugly and not scalable, we have faced issues with it in the past. It needs to be generic enough to work with all sort of parameters but I understand that the way we are doing it right now (by reflection etc.) we will never be able to do it in a generic fashion and it might not even be possible to use it if someone uses advance features like custom beans etc. However, fixing it will be too much to accommodate in this JIRA, we can create another JIRA to address this issue.

:+1:  for this change from my side.

pavankumar526: Agreed, +1 for the patch. @PraveenAdlakha  Please raise a jira to fix this, this code looks clumpsy.

PraveenAdlakha: @pavankumar526  We already have a jira for this : FALCON-1098

ajayyadava:
Will commit shortly.

PR #89, opened 2016-04-08
FALCON-1817 : Extend xsd for Spark execution engine

peeyushb:
Pull request is available that will extend the cluster and process XSD to include the spark execution engine.

### DIFF
--- a/client/src/main/resources/process-0.1.xsd
+++ b/client/src/main/resources/process-0.1.xsd
@@ -349,7 +349,6 @@
                     master element is optional and indicates the url of the Spark Master. Ex: yarn-client, yarn-master, or local. This will override the Spark interface, if define in cluster entity.
                     mode element if present indicates the mode of spark, where to run spark driver program. Ex: client,cluster.
                     name element indicates the name of the spark application.
-                    application-type indicates type of the spark application java, python or spark-sql.
                     class element if present, indicates the spark's application main class.
                     jar element indicates a comma separated list of jars or python files. Also provide the jar or file directory path to workflow attribute path
                     spark-opts element if present, contains a list of spark options that can be passed to spark driver.
--- a/client/src/main/resources/process-0.1.xsd
+++ b/client/src/main/resources/process-0.1.xsd
@@ -358,7 +358,7 @@
         <xs:sequence>
             <xs:element name="master" type="xs:string" minOccurs="0" maxOccurs="1"/>
             <xs:element name="mode" type="xs:string" minOccurs="0" maxOccurs="1"/>
-            <xs:element name="name" type="xs:string" minOccurs="0" maxOccurs="1"/>
+            <xs:element name="name" type="xs:string" minOccurs="1" maxOccurs="1"/>
             <xs:element name="class" type="xs:string" minOccurs="0" maxOccurs="1"/>
             <xs:element name="jar" type="xs:string" minOccurs="1" maxOccurs="1"/>
             <xs:element name="spark-opts" type="xs:string" minOccurs="0" maxOccurs="1"/>
--- a/client/src/main/resources/process-0.1.xsd
+++ b/client/src/main/resources/process-0.1.xsd
@@ -152,6 +152,13 @@
                     </xs:documentation>
                 </xs:annotation>
             </xs:element>
+            <xs:element type="spark-attributes" name="spark-attributes" minOccurs="0">
+                <xs:annotation>
+                    <xs:documentation>Defines the spark related attributes to run the Spark application. Attributes
+                        must be defined when workflow engine "spark" is selected.
+                    </xs:documentation>
+                </xs:annotation>
+            </xs:element>
             <xs:element type="retry" name="retry" minOccurs="0">
                 <xs:annotation>
                     <xs:documentation>Retry defines how to handle workflow failures. Policy type - periodic, exponential
@@ -331,8 +338,34 @@
             <xs:enumeration value="oozie"/>
             <xs:enumeration value="pig"/>
             <xs:enumeration value="hive"/>
+            <xs:enumeration value="spark"/>
         </xs:restriction>
     </xs:simpleType>
+
+    <xs:complexType name="spark-attributes">
+        <xs:annotation>
+            <xs:documentation>
+                List of Spark attributes to run Spark application:
+                    master element is optional and indicates the url of the Spark Master. Ex: yarn-client, yarn-master, or local. This will override the Spark interface, if define in cluster entity.
+                    mode element if present indicates the mode of spark, where to run spark driver program. Ex: client,cluster.
+                    name element indicates the name of the spark application.
+                    application-type indicates type of the spark application java, python or spark-sql.
+                    class element if present, indicates the spark's application main class.
+                    jar element indicates a comma separated list of jars or python files. Also provide the jar or file directory path to workflow attribute path
+                    spark-opts element if present, contains a list of spark options that can be passed to spark driver.
+                    arg element if present, contains arguments that can be passed to spark application.
+            </xs:documentation>
+        </xs:annotation>
+        <xs:sequence>
+            <xs:element name="master" type="xs:string" minOccurs="0" maxOccurs="1"/>
+            <xs:element name="mode" type="xs:string" minOccurs="0" maxOccurs="1"/>
+            <xs:element name="name" type="xs:string" minOccurs="0" maxOccurs="1"/>
+            <xs:element name="class" type="xs:string" minOccurs="0" maxOccurs="1"/>
+            <xs:element name="jar" type="xs:string" minOccurs="1" maxOccurs="1"/>
+            <xs:element name="spark-opts" type="xs:string" minOccurs="0" maxOccurs="1"/>
+            <xs:element name="arg" type="xs:string" minOccurs="0" maxOccurs="unbounded"/>
+        </xs:sequence>
+    </xs:complexType>
 
     <xs:complexType name="retry">
         <xs:attribute type="policy-type" name="policy" use="required"/>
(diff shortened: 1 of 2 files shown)


client/src/main/resources/process-0.1.xsd
```diff
@@ -331,6 +338,40 @@
...
+                    name element indicates the name of the spark application.
+                    application-type indicates type of the spark application java, python.
+                    class element if present, indicates the spark's application main class.
+                    jar element indicates a comma separated list of jars or python files.
+                    spark-opts element if present, contains a list of spark options that can be passed to spark driver.
+                    arg element if present, contains arguments that can be passed to spark application.
+            </xs:documentation>
+        </xs:annotation>
+        <xs:sequence>
+            <xs:element name="master" type="xs:string" minOccurs="1" maxOccurs="1"/>
```
pallavi-rao: Some of these might be useful to define at the cluster level, example, master, mode. The current cluster xsd may need to be extended to allow for additional properties for an interface.

peeyushb: Initially I have thought to put some of the attributes at cluster level but it does not look easy to extend the cluster xsd with the set of interface properties as here. So I have defined all the required spark attributes here in process xsd.

pallavi-rao: The reason a cluster definition exists is for re-use of such attributes, so user doesn't have to repeat it in every process. I think we should change cluster xsd if required and not necessarily stick to existing interface definition. May be add optional properties to the interface element? Or, add this to the list of custom properties (as we do for hive now)?

peeyushb: Spark gives flexibility to the user to run application by specifying any value with master option like local, yarn-client, yarn-cluster or spark:// url . So if value to master specified through cluster xsd, then user will not be able to change the value of master option easily with different type of Spark application defined in process xsd.

client/src/main/resources/process-0.1.xsd
```diff
@@ -331,6 +338,40 @@
...
+            <xs:element name="jar" type="xs:string" minOccurs="1" maxOccurs="1"/>
+            <xs:element name="spark-opts" type="xs:string" minOccurs="0" maxOccurs="1"/>
+            <xs:element name="arg" type="xs:string" minOccurs="0" maxOccurs="unbounded"/>
+        </xs:sequence>
+    </xs:complexType>
+
+    <xs:simpleType name="application-type">
+        <xs:restriction base="xs:string">
+            <xs:enumeration value="java"/>
+            <xs:enumeration value="python"/>
```
vramachan: Just need a clarification. where and how are the application-type is used? I see python and java. Spark also supports Scala and R bindings.

client/src/main/resources/process-0.1.xsd
```diff
@@ -322,7 +329,7 @@
         <xs:attribute type="xs:string" name="name" use="optional"/>
         <xs:attribute type="xs:string" name="version" use="optional" default="1.0"/>
         <xs:attribute type="engine-type" name="engine" use="optional" default="oozie"/>
-        <xs:attribute type="xs:string" name="path" use="required"/>
+        <xs:attribute type="xs:string" name="path" use="optional"/>
```
vrangan: Wouldn't this be inconsistent with other engines.   Can we keep it as required and handle this in our code?

client/src/main/resources/process-0.1.xsd
```diff
@@ -331,6 +338,40 @@
...
+                    jar element indicates a comma separated list of jars or python files.
+                    spark-opts element if present, contains a list of spark options that can be passed to spark driver.
+                    arg element if present, contains arguments that can be passed to spark application.
+            </xs:documentation>
+        </xs:annotation>
+        <xs:sequence>
+            <xs:element name="master" type="xs:string" minOccurs="1" maxOccurs="1"/>
+            <xs:element name="mode" type="xs:string" minOccurs="0" maxOccurs="1"/>
+            <xs:element name="name" type="xs:string" minOccurs="1" maxOccurs="1"/>
+            <xs:element name="application-type" type="application-type"  minOccurs="0" maxOccurs="1" default="java"/>
```
vrangan: Let us remove application-type for now.   We are not really using this right now.  We can add this if a need arises later.   My worry is we introduce something which we never use and this adds to confusion like the version in the interface definitions in the cluster entity.

client/src/main/resources/process-0.1.xsd
vrangan: I think it may be good to have master in the cluster spark interface and overridable using the process properties as needed

peeyushb: I have define the master in cluster xsd as well. If user will define master in process xsd then it will take precedence than master in cluster xsd. But if no master in process xsd then master from cluster xsd will be used to run spark application.

client/src/main/resources/process-0.1.xsd
peeyushb: I defined the application-type to handle more specific spark application attributes. But removed now considering only support is for Java/Python only.

client/src/main/resources/process-0.1.xsd
peeyushb: I have keep it "required" only and will ask user to specify spark application(jar/python) directory path with Process entity workflow path.

client/src/main/resources/process-0.1.xsd
peeyushb: Removed.

client/src/main/resources/process-0.1.xsd
```diff
@@ -331,9 +338,35 @@
...
     </xs:simpleType>
 
+    <xs:complexType name="spark-attributes">
+        <xs:annotation>
+            <xs:documentation>
+                List of Spark attributes to run Spark application:
+                    master element is optional and indicates the url of the Spark Master. Ex: yarn-client, yarn-master, or local. This will override the Spark interface, if define in cluster entity.
+                    mode element if present indicates the mode of spark, where to run spark driver program. Ex: client,cluster.
+                    name element indicates the name of the spark application.
+                    application-type indicates type of the spark application java, python or spark-sql.
```
pallavi-rao: Remove this line

client/src/main/resources/process-0.1.xsd
```diff
@@ -331,9 +338,35 @@
...
+                    class element if present, indicates the spark's application main class.
+                    jar element indicates a comma separated list of jars or python files. Also provide the jar or file directory path to workflow attribute path
```
pallavi-rao: Should you still say python?

pallavi-rao:
+1 apart from the minor nits

client/src/main/resources/process-0.1.xsd
peeyushb: @pallavi-rao  If user want to run python spark application, then they must specify with <jar> file option. It is as per Oozie Spark xsd. Ideally I prefer to use <file> instead of <jar>.

client/src/main/resources/process-0.1.xsd
```diff
@@ -331,9 +338,34 @@
...
+                    spark-opts element if present, contains a list of spark options that can be passed to spark driver.
+                    arg element if present, contains arguments that can be passed to spark application.
+            </xs:documentation>
+        </xs:annotation>
+        <xs:sequence>
+            <xs:element name="master" type="xs:string" minOccurs="0" maxOccurs="1"/>
+            <xs:element name="mode" type="xs:string" minOccurs="0" maxOccurs="1"/>
+            <xs:element name="name" type="xs:string" minOccurs="0" maxOccurs="1"/>
```
pavankumar526: shouldn't we made name,  class also require arguments ? Why these are optional ?

client/src/main/resources/process-0.1.xsd
```diff
@@ -331,9 +338,34 @@
...
+                    mode element if present indicates the mode of spark, where to run spark driver program. Ex: client,cluster.
+                    name element indicates the name of the spark application.
+                    class element if present, indicates the spark's application main class.
+                    jar element indicates a comma separated list of jars or python files. Also provide the jar or file directory path to workflow attribute path
+                    spark-opts element if present, contains a list of spark options that can be passed to spark driver.
+                    arg element if present, contains arguments that can be passed to spark application.
+            </xs:documentation>
+        </xs:annotation>
+        <xs:sequence>
+            <xs:element name="master" type="xs:string" minOccurs="0" maxOccurs="1"/>
```
pavankumar526: We should handle the case where master is not mentioned in both cluster definition and as well as in process definition

client/src/main/resources/process-0.1.xsd
```diff
@@ -331,9 +338,34 @@
...
+            <xs:element name="mode" type="xs:string" minOccurs="0" maxOccurs="1"/>
```
pavankumar526: Mode is required in case of jobs submitted to spark cluster

client/src/main/resources/process-0.1.xsd
peeyushb: name is made optional because there are chances that Spark application will itself have app name using API setAppName(). So it is upto the user wish to specify name in entity or in application itself. 

class is made optional to handle the Python Spark application, which can only be specified as path with "jar" option.

client/src/main/resources/process-0.1.xsd
peeyushb: Yes, we will have a check for this in implementation. If Spark master is not specified either in cluster or in process entity, an exception will be thrown.

client/src/main/resources/process-0.1.xsd
peeyushb: If master is based on yarn i.e yarn-client or yarn-cluster , then by default mode will be client and cluster respectively. But if master is spark:// url , then user can specify the mode to be client or cluster and also we can think to make the mode default to client.

peeyushb: @pavankumar526  I have made name parameter as required, so that user must specify app name with process entity. This will help if in case the user could not able to provide or forgot to mention app name with in the Spark application.

pavankumar526: Ok thanks

pavankumar526:
Thanks @peeyushb  for adressing comments. Patch looks good. +1

peeyushb:
Thanks @pavankumar526 for reviewing the patch. If no further comments, I will commit this patch soon.

PR #93, opened 2016-04-12
FALCON-1335 Backend support of instance search of a group of entities

yzheng-hortonworks:
Tested REST API: api/instance/search


bvellanki:
@yzheng-hortonworks  +1 for the patch. Please add IT tests for instance search of a group of entities. This should be possible under MetadataResourceJerseyIT.

common/src/main/java/org/apache/falcon/metadata/GraphUtils.java
```diff
@@ -81,4 +87,51 @@ public static String edgeString(final Edge edge) {
...
+        LOG.debug("Finding vertex for: name={}, type={}", name, type);
+        RelationshipType relationShipType;
+        switch (type) {
+        case "FEED":
+            relationShipType = RelationshipType.FEED_ENTITY;
+            break;
+        case "PROCESS":
+            relationShipType = RelationshipType.PROCESS_ENTITY;
+            break;
+        default:
```
vramachan: It appears that GraphUtils.findVertex() is a public utility method. It's legitimate to get relationshipTypes other than FEED and PROCESS as defined in the RelationshipType.java. if this method is used by this feature, it needs to be renamed or moved to some place else.

webapp/src/main/java/org/apache/falcon/resource/InstanceManager.java
```diff
@@ -306,6 +307,28 @@ public InstancesResult reRunInstance(
...
+            @DefaultValue("") @QueryParam("nameseq") String nameSubsequence,
+            @DefaultValue("") @QueryParam("tagkeys") String tagKeywords,
+            @DefaultValue("") @QueryParam("start") String nominalStartTime,
+            @DefaultValue("") @QueryParam("end") String nominalEndTime,
+            @DefaultValue("") @QueryParam("instanceStatus") String status,
+            @DefaultValue("") @QueryParam("orderBy") String orderBy,
+            @DefaultValue("0") @QueryParam("offset") Integer offset,
+            @QueryParam("numResults") Integer resultsPerPage) {
+        type = StringUtils.isEmpty(type) ? "feed,process" : type;
+        resultsPerPage = resultsPerPage == null ? getDefaultResultsPerPage() : resultsPerPage;
```
vramachan: double assignment, typo?

common/src/main/java/org/apache/falcon/metadata/GraphUtils.java
yzheng-hortonworks: Good point, Venky. Fixed.

client/src/main/java/org/apache/falcon/metadata/RelationshipType.java
```diff
@@ -64,4 +64,15 @@ public static RelationshipType fromString(String value) {
 
         throw new IllegalArgumentException("No constant with value " + value + " found");
     }
+
+    public static RelationshipType fromSchedulableEntityType(String type) {
+        switch (type) {
+        case "FEED":
```
sowmyaramesh: Please use EntityType.getEnum(type) instead of using string literals

client/src/main/java/org/apache/falcon/metadata/RelationshipType.java
```diff
@@ -64,4 +64,15 @@ public static RelationshipType fromString(String value) {
...
+            return RelationshipType.FEED_ENTITY;
+        case "PROCESS":
```
sowmyaramesh: Same as above

prism/src/main/java/org/apache/falcon/resource/AbstractInstanceManager.java
```diff
@@ -676,6 +683,79 @@ public TriageResult triageInstance(String entityType, String entityName, String
         }
     }
 
+    public InstancesResult searchInstances(String type, String nameSubsequence, String tagKeywords,
```
peeyushb: It looks like that searchInstances function is doing multiple things. Is it possible to split the function to reflect more appropriately.

yzheng-hortonworks: Could you elaborate your thoughts? This function retrieves relevant entities and then make a batch query on their instances on Titan DB.

peeyushb: searchInstances function is retrieving and populating entityVertices, adding and running query to get instanceMap, populating instances List with search result. These tasks are happening in searchInstances function along with calling of sortInstancesPagination() and returning InstanceResult as output.  So I thought it is better to split the searchInstances function with mentioned tasks in subfunction for better manageability going forward.

client/src/main/java/org/apache/falcon/metadata/RelationshipType.java
yzheng-hortonworks: Changed to EntityType

client/src/main/java/org/apache/falcon/metadata/RelationshipType.java
yzheng-hortonworks: Changed to EntityType

prism/src/main/java/org/apache/falcon/resource/AbstractInstanceManager.java
yzheng-hortonworks: @peeyushb Reorganized the code

peeyushb:
Looks good to me. +1.

peeyushb:
For this pull request, @yzheng-hortonworks mentioned that she need to add unit tests. So she will update this pull request.

yzheng-hortonworks:
Accidentally deleted the branch... Created a new pull request: https://github.com/apache/falcon/pull/106

PR #95, opened 2016-04-12
FALCON-1896 Failure in Falcon build in distro module

PraveenAdlakha:


bvellanki:
+1 works for me. Will commit shortly

PR #96, opened 2016-04-13
FALCON-1898 Build fails on assembly:assembly when using maven-assembly-plugin version 2.6

bvellanki:


vrangan:
Thanks for catching it @bvellanki .   +1

vramachan:
LGTM +1

bvellanki:
Thank you for review @vrangan  and @vramachan. I will commit this to unblock jenkins builds. If there is a cleaner way to solve the error, we will revisit the issue in future.

PR #103, opened 2016-04-16
FALCON-1897 Extension Job Management: CLI support

yzheng-hortonworks:
Note: There is a bug to run Falcon CLI after the code refactoring. @sowmyaramesh has the patch to fix this and will send a pull request. Tested all the CLI methods with her fix patch.


bvellanki:
👍  for the code Ying, please add a few positive/negative tests in FalconCLIIT and I will commit the patch.

yzheng-hortonworks:
@bvellanki Thanks for review. Since there are 10 rest apis and 10 cli commands, the IT tests for all of them will be handled in a separate jira: https://issues.apache.org/jira/browse/FALCON-1905. I'll add both positive and negative tests there. Thanks for your reminder.

bvellanki:
That sounds good to me. If there are no other comments from community, I will commit the patch tomorrow.

PR #111, opened 2016-04-21
FALCON-1916: Allow RM principal to be specified in Cluster entity

vrangan:
When we have clusters setup with individual security domains with  cross realm trusts, the RM principal needs to be specified explicitly to disambiguate the realm of the resource manager of the corresponding cluster entities.

This patch allows RM principal to be set and processed from the cluster properties


common/src/main/java/org/apache/falcon/hadoop/HadoopClientFactory.java
```diff
@@ -214,11 +215,19 @@ public FileSystem run() throws Exception {
...
         jobConf.set(MR_JT_ADDRESS_KEY, executeUrl);
         jobConf.set(YARN_RM_ADDRESS_KEY, executeUrl);
-
+        /**
+         * It is possible that the RM/JT principal can be different between clusters,
+         * for example, the cluster is using a different KDC with cross-domain trust
+         * with the Falcon KDC.   in that case, we want to allow the user to provide
+         * the RM principal similar to NN principal.
+         */
+        if (UserGroupInformation.isSecurityEnabled() && StringUtils.isNotEmpty(rmPrincipal)) {
```
vramachan: Usually StringUtils.isNotBlank() is used to make sure no empty string passed in for the property.

vramachan: +1 
Looks good with one minor point. 
I believe we can't/don't have tests for security related aspects?

bvellanki:
👍 Venkat, will commit shortly.

PR #114, opened 2016-04-22
FALCON-1906 upgrade falcon to java 8

PraveenAdlakha:


peeyushb:
@PraveenAdlakha There are Falcon modules where source and target version is 1.7 for "maven-compiler-plugin". Can you just check other modules to upgrade to 1.8.

PraveenAdlakha:
@peeyushb  have made the changes please review.

bvellanki:
👍 Looks good to me. Please comment on end2end testing done for this Jira.

vrangan:
Why do you want to disable JDK 7 support?

PraveenAdlakha:
@vrangan Right now we are not using any code which is specific to java 8 so should not cause any problem,we can do testing for the same.

But we should migrate to java 8 as java 7 already had its end of life.Any suggestions that you have are welcome.

vrangan:
There are  customers that still use JDK 7 as the platforms.     Building with JDK 7 does not preclude use of JDK 8 as the runtime platform.   So, we should defer until there is a real need to change the source compatibility to JDK 8

PR #116, opened 2016-04-26
FALCON-1623 Implement safemode in Falcon Server

bvellanki:


sowmyaramesh:
I have couple of comments for design doc. Sorry, I should have reviewed it when doc was provided long time. 

1> Can you talk about behavior for safe mode when the authorization is enabled & disabled?
2> Design doc says "Super user can perform update operation only on cluster entities". Shouldn't this be set safemode ON/OFF instead of cluster updates? 
3> Even the admin users/groups should be able to enable/disable safemode and not only super user
4> Doc mentions "Entity operations not allowed are : update, schedule, touch, delete, submit,
submitAndSchedule". Did you miss resume or is it allowed as its not allowed for instances.
5> In section "UI support for safemode" doesn't have the updated list of operations disallowed. Can you please update it?

Again I am sorry for the late review. Thanks!

sowmyaramesh:
I just started reviewing the code, not fully done with it. 
Please add "safemode" in the below code in  DefaultAuthorizationProvider.authorizeResource else it will not work for admin users/groups

```
            if ("admin".equals(resource)) {
                if (!("version".equals(action) || "clearuser".equals(action) || "getuser".equals(action))) {
                    authorizeAdminResource(authenticatedUGI, action);
                }
```

bvellanki:
@sowmyaramesh  : The above suggestion is inaccurate. The code says "allow non-admin users to access version, clearuser and getuser". So we should not add safemode to this code.

sowmyaramesh:
@bvellanki : Where does it say that? If you look at authorizeAdminResource it is what validates if the user has admin privileges. 

I don't think current code works for admin user. Did you test it for admin user not super user case?

bvellanki:
@sowmyaramesh Please look at the code again. 
` if (isSuperUser(authenticatedUGI)) {
                return;
            }`

This part would let a super user perform setsafemode action. In DefaultAuthorizationProvider.authorizeAdminResource you have ` if (adminUsers.contains(authenticatedUser) || isUserInAdminGroups(authenticatedUGI)) {
            return;
        }`

This part allows admin users to perform setsafemode action. Everyone else is denied permission to perform setsafemode action

sowmyaramesh:
@bvellanki : My bad, missed the ! in the if statement

```
if ("admin".equals(resource)) {
                if (!("version".equals(action) || "clearuser".equals(action) || "getuser".equals(action))) {
                    authorizeAdminResource(authenticatedUGI, action);
                }
```

bvellanki:
@sowmyaramesh  : Updated the design document in https://issues.apache.org/jira/browse/FALCON-1623

prism/src/main/java/org/apache/falcon/FalconServer.java
```diff
@@ -88,6 +94,14 @@ public static void main(String[] args) throws Exception {
             appPath = cmd.getOptionValue(APP_PATH);
         }
 
+        String isSafeMode = "false";
+        if (cmd.hasOption(SAFE_MODE)) {
+            isSafeMode = cmd.getOptionValue(SAFE_MODE);
+            validateSafemode(isSafeMode);
+        }
+        LOG.info("Setting system property Safemode to : {}", isSafeMode);
+        System.setProperty(StartupProperties.SAFEMODE_PROPERTY, isSafeMode);
```
vrangan: @bvellanki    I am wondering whether we need a bit stronger mechanism for keeping track of  safemode internally.   For example, if someone adds the java system property to the Java command line, we will be in safe mode.   Wondering whether we can track it by tracking it in a system file - something that is persistent also and enables safemode to survive restarts as a potential enhancement

bvellanki: @vrangan  : As part of design doc, we said that restart will get FalconServer out of safemode. We should only consider persisting the property if we want safemode to survive restarts. If we want to make such a change in design, I request we file it as enhancement Jira. 

StartupProperties are not initialized until after falcon server is started. So if we want to use the command line args for start script to set safemode, the only way to pass that on to server is via System property. Please let me know if you have any other recommendation.

bvellanki:
@sowmyaramesh @vrangan  I updated patch to be able to persist safemode setting between restarts of falcon server. This is done based on your feedback. Please review.

yzheng-hortonworks:
Patch looks good. One general comment: You use a file to persist whether Falcon server is in safe mode or not. For long term use, looks better to have a property file owned by Falcon to persist important runtime properties (could be a separate jira) :)

vrangan:
+1. Thanks

PR #117, opened 2016-04-26
FALCON-1909 Catalog instance triage action fails with null pointer ex…

bvellanki:
…ception.

### DIFF
--- a/common/src/main/java/org/apache/falcon/entity/CatalogStorage.java
+++ b/common/src/main/java/org/apache/falcon/entity/CatalogStorage.java
@@ -90,7 +90,6 @@ public class CatalogStorage extends Configured implements Storage {
 
     public CatalogStorage(Cluster cluster, CatalogTable table) throws URISyntaxException {
         this(ClusterHelper.getInterface(cluster, Interfacetype.REGISTRY).getEndpoint(), table);
-        verifyAndUpdateConfiguration(getConf());
     }
 
     protected CatalogStorage(String catalogUrl, CatalogTable table) throws URISyntaxException {
@@ -110,6 +109,8 @@ public class CatalogStorage extends Configured implements Storage {
     private void verifyAndUpdateConfiguration(Configuration conf) {
         if (conf == null) {
             setConf(new Configuration());
+        } else {
+            setConf(conf);
         }
     }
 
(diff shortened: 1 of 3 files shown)


common/src/main/java/org/apache/falcon/entity/CatalogStorage.java
```diff
@@ -90,6 +90,7 @@ protected CatalogStorage(Feed feed) throws URISyntaxException {
 
     public CatalogStorage(Cluster cluster, CatalogTable table) throws URISyntaxException {
         this(ClusterHelper.getInterface(cluster, Interfacetype.REGISTRY).getEndpoint(), table);
+        verifyAndUpdateConfiguration(getConf());
```
yzheng-hortonworks: You don't need it here. verifyAndUpdateConfiguration is already called in CatalogStorage(String catalogUrl, CatalogTable table).

common/src/main/java/org/apache/falcon/entity/CatalogStorage.java
```diff
@@ -186,13 +193,14 @@ protected CatalogStorage(String uriTemplate) throws URISyntaxException {
...
         this.catalogUrl = uri.getScheme() + "://" + uri.getAuthority();
-
+        verifyAndUpdateConfiguration(getConf());
         parseUriTemplate(uri);
     }
 
     protected CatalogStorage(String uriTemplate, Configuration conf) throws URISyntaxException {
         this(uriTemplate);
         setConf(conf);
+        verifyAndUpdateConfiguration(conf);
```
yzheng-hortonworks: You don't need it here. verifyAndUpdateConfiguration is already called in CatalogStorage(String uriTemplate).

bvellanki: @yzheng-hortonworks  You are correct in both comments, I was overzealous, will submit a new pull request. Thanks for catching it.

common/src/main/java/org/apache/falcon/entity/CatalogStorage.java
```diff
@@ -100,12 +100,20 @@ protected CatalogStorage(String catalogUrl, String tableUri) throws URISyntaxExc
...
             throw new IllegalArgumentException("Catalog Registry URL cannot be null or empty");
         }
-
+        verifyAndUpdateConfiguration(getConf());
         this.catalogUrl = catalogUrl;
 
         parseFeedUri(tableUri);
     }
 
+    private void verifyAndUpdateConfiguration(Configuration conf) {
```
yzheng-hortonworks: Could you move this to be under all the CatalogStorage functions?

yzheng-hortonworks: I assume you are always using getConf() as the input argument? Then, a better way is to have verifyAndUpdateConfiguration() with no input and call getConf() in the function, so you don't need to reset configuration with the same value.

yzheng-hortonworks:
+1

vrangan:
+1

PraveenAdlakha:
+1 LGTM

PR #119, opened 2016-04-28
FALCON-1926 : Filter out effectively non-falcon related JMS messages …

vramachan:
* Falcon to retrieve Oozie JMS notifications that belong to Falcon generated Workflows .. ignoring notifications generated by other WF directly submitted to Oozie.

### DIFF
--- a/messaging/src/main/java/org/apache/falcon/messaging/JMSMessageConsumer.java
+++ b/messaging/src/main/java/org/apache/falcon/messaging/JMSMessageConsumer.java
@@ -92,7 +92,8 @@ public class JMSMessageConsumer implements MessageListener, ExceptionListener {
 
             topicSession = (TopicSession) connection.createSession(false, Session.AUTO_ACKNOWLEDGE);
             Topic destination = topicSession.createTopic(topicName);
-            topicSubscriber = topicSession.createDurableSubscriber(destination, FALCON_CLIENT_ID);
+            topicSubscriber = topicSession.createDurableSubscriber(destination, FALCON_CLIENT_ID,
+                    WorkflowNameBuilder.WorkflowName.getJMSFalconSelector(), false);
             topicSubscriber.setMessageListener(this);
 
             connection.setExceptionListener(this);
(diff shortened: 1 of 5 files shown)


messaging/src/main/java/org/apache/falcon/messaging/JMSMessageConsumer.java
```diff
@@ -92,7 +92,8 @@ public void startSubscriber() throws FalconException {
 
             topicSession = (TopicSession) connection.createSession(false, Session.AUTO_ACKNOWLEDGE);
             Topic destination = topicSession.createTopic(topicName);
-            topicSubscriber = topicSession.createDurableSubscriber(destination, FALCON_CLIENT_ID);
+            topicSubscriber = topicSession.createDurableSubscriber(destination, FALCON_CLIENT_ID,
```
pallavi-rao: With this change, any message from post-processing will get filtered. We should clearly call it out as a backward-incompatible change/removed feature.

Also, if this change is in place, we can also remove some of the code that allows processing of messages from both post-processing action (which was done for backward compatibility) and Oozie JMS notification.

vramachan: @pallavi-rao post-processing now sends only user notifications (the code in postprocessing to send system notification to FALCON.ENTITY.TOPIC is removed since Oozie JMS covers that). In that regard, I think it is a backward compatible  issue -- Users need to enable Oozie JMS notification. This is anyways needed for Instance Search to work.

Regarding the second one, are you referring to the TextMessage and MapMessage cases in the code? MapMessage in the JMS Consumer reads messages from post processing while TextMessage represents the Oozie JMS messages. Can you please confirm?

bvellanki:
Based on response given by @vramachan , the code looks good and does what is expected. Please update the documentation as well to say that user should enable Oozie JMS notification. Once documentation is provided the code looks good.

yzheng-hortonworks:
+1

vramachan:
@bvellanki Added docs in the configuration section on the need for oozie JMS. There is a separate JIRA FALCON-1948 to detail out configuring Oozie for Falcon. Will address that shortly.

bvellanki:
+1 will commit shortly.

PR #122, opened 2016-05-03
FALCON - 1930 : testProcessInstanceRerunFailedWorkflowAction failed with NULLPOINTEREXCEPTION

muraliramasami:


falcon-regression/merlin-core/src/main/java/org/apache/falcon/regression/core/util/OozieUtil.java
```diff
@@ -714,7 +714,9 @@ public static String getActionStatus(OozieClient oozieClient, String workflowId,
         List<WorkflowAction> wfAction = oozieClient.getJobInfo(workflowId).getActions();
         for (WorkflowAction wf : wfAction) {
             if (wf.getName().contains(actionName)) {
-                return wf.getExternalStatus();
+                if (wf.getExternalStatus() != null) {
```
PraveenAdlakha: please use 'and' operator in if condition and combine both the if statements for better readability.

muraliramasami: @PraveenAdlakha  Thanks for the review. I have modified the code.

falcon-regression/merlin-core/src/main/java/org/apache/falcon/regression/core/util/OozieUtil.java
```diff
@@ -713,7 +713,7 @@ public static String getActionStatus(OozieClient oozieClient, String workflowId,
         throws OozieClientException {
         List<WorkflowAction> wfAction = oozieClient.getJobInfo(workflowId).getActions();
         for (WorkflowAction wf : wfAction) {
-            if (wf.getName().contains(actionName)) {
+            if ((wf.getName().contains(actionName)) && (wf.getExternalStatus() != null)) {
```
raghavgautam: Do we know why we got null in the first place ?

PR #123, opened 2016-05-03
FALCON-1931 : Adding multiCluster tag for Multiple Cluster scenarios

muraliramasami:
In the regression test suite, there are lot of places multiCluster annotation has been defined to denote it uses multiple cluster. It is very useful annotation since we require additional cluster and setup to test the test case . Some of the major test cases such as feed replication and Hive DR are missing this annotation.

### DIFF
--- a/falcon-regression/merlin/src/test/java/org/apache/falcon/regression/hcat/HCatFeedOperationsTest.java
+++ b/falcon-regression/merlin/src/test/java/org/apache/falcon/regression/hcat/HCatFeedOperationsTest.java
@@ -119,7 +119,7 @@ public class HCatFeedOperationsTest extends BaseTestClass {
      *
      * @throws Exception
      */
-    @Test(groups = {"singleCluster"})
+    @Test(groups = {"multiCluster"})
     public void submitFeedWhenTableDoesNotExist() throws Exception {
         Bundle.submitCluster(bundles[1]);
         feed = bundles[1].getInputFeedFromBundle();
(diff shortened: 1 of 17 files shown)


falcon-regression/merlin/src/test/java/org/apache/falcon/regression/hcat/HCatFeedOperationsTest.java
```diff
@@ -119,7 +119,7 @@ public void tearDownClass() throws IOException {
      *
      * @throws Exception
      */
-    @Test(groups = {"singleCluster"})
+    @Test(groups = {"multiCluster"})
```
raghavgautam: Why is this a multiCluster test ?

raghavgautam:
@muraliramasami Please answer the question rest everything looks good.

falcon-regression/merlin/src/test/java/org/apache/falcon/regression/hcat/HCatFeedOperationsTest.java
muraliramasami: @raghavgautam  submitFeedWhenTableDoesNotExist uses second cluster information to submit the feed. Since we are separating the execution of single vs multi cluster test cases, this will come under multiCluster tag.

muraliramasami:
@raghavgautam  Thank you for reviewing the code.

raghavgautam:
👍

PR #124, opened 2016-05-03
FALCON-1932 Extension CLI should support common options

yzheng-hortonworks:
Tested CLI command with common options.

### DIFF
--- a/cli/src/main/java/org/apache/falcon/cli/FalconExtensionCLI.java
+++ b/cli/src/main/java/org/apache/falcon/cli/FalconExtensionCLI.java
@@ -167,6 +167,10 @@ public class FalconExtensionCLI {
         group.addOption(delete);
         extensionOptions.addOptionGroup(group);
 
+        Option url = new Option(FalconCLIConstants.URL_OPTION, true, "Falcon URL");
+        Option doAs = new Option(FalconCLIConstants.DO_AS_OPT, true, "doAs user");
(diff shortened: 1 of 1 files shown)


bvellanki:
+1

yzheng-hortonworks:
@sowmyaramesh To follow up with our talk, our current CLI framework doesn't check whether an option is used in a command or not, i.e. they are all optional. In other words, whether I add an option or not doesn't change your extension repository results, if you don't use these options. For example, "bin/falcon extension -enumerate -url http://localhost:15000 -d -ddd " gives the same result as "bin/falcon extension -enumerate -url http://localhost:15000".

This resembles with REST api design, i.e. it won't throw error if user pass an unrecognized parameter. Hope this resolves your concern.

If you want to be able to throw error for unsupported options, you need to change our CLI framework.

sowmyaramesh:
@yzheng-hortonworks : There are two ways of handling it.
1> Instead of adding by default to Extension CLI it can be added only to commands which honor it. That way by default if unsupported option is passed it will throw exception. This is how it is handled today in the code
2> Other way is to handle in specific commands

As your code is in accordance with the current design I will give a ship ti. We can revisit this if required in the future. Thanks!

sowmyaramesh:
+1

yzheng-hortonworks:
Thanks @sowmyaramesh . Yes, and this would be a lot of work, since every CLI command only validates required parameters but not unsupported ones. We need to change every CLI command to make it happen.

PR #127, opened 2016-05-05
FALCON-1085 Support Cluster entity updates in Falcon Server

bvellanki:
Added basic documentation, https://issues.apache.org/jira/browse/FALCON-1937 will contain detailed documentation.

### DIFF
--- a/common/src/main/java/org/apache/falcon/entity/ClusterHelper.java
+++ b/common/src/main/java/org/apache/falcon/entity/ClusterHelper.java
@@ -32,6 +32,8 @@ import org.apache.falcon.hadoop.HadoopClientFactory;
 import org.apache.falcon.security.SecurityUtil;
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.fs.Path;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import java.util.HashMap;
 import java.util.List;
@@ -45,7 +47,7 @@ public final class ClusterHelper {
     public static final String WORKINGDIR = "working";
     public static final String NO_USER_BROKER_URL = "NA";
     public static final String EMPTY_DIR_NAME = "EMPTY_DIR_DONT_DELETE";
-
+    private static final Logger LOG = LoggerFactory.getLogger(ClusterHelper.class);
 
     private ClusterHelper() {
     }
@@ -123,6 +125,9 @@ public final class ClusterHelper {
     }
 
     public static Interface getInterface(Cluster cluster, Interfacetype type) {
+        if (cluster.getInterfaces() == null) {
+            return null;
+        }
         for (Interface interf : cluster.getInterfaces().getInterfaces()) {
             if (interf.getType() == type) {
                 return interf;
@@ -143,6 +148,9 @@ public final class ClusterHelper {
 
 
     public static Location getLocation(Cluster cluster, ClusterLocationType clusterLocationType) {
+        if (cluster.getLocations() == null) {
+            return null;
+        }
         for (Location loc : cluster.getLocations().getLocations()) {
             if (loc.getName().equals(clusterLocationType)) {
                 return loc;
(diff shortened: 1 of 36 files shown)


webapp/src/test/java/org/apache/falcon/cli/FalconClusterUpdateCLIIT.java
```diff
@@ -0,0 +1,141 @@
...
+        TestContext.deleteEntitiesFromStore();
+    }
+
+
+    public void testUpdateClusterCommands() throws Exception {
+
+        FalconCLI.OUT.set(stream);
+
+        String filePath;
+        TestContext context = new TestContext();
```
peeyushb: "context" has been defined again but it is defined as attribute already.

webapp/src/test/java/org/apache/falcon/cli/FalconClusterUpdateCLIIT.java
```diff
@@ -0,0 +1,141 @@
...
+        Map<String, String> overlay = context.getUniqueOverlay();
```
peeyushb: "overlay" is declared as attribute already.

prism/src/main/java/org/apache/falcon/resource/AbstractEntityManager.java
```diff
@@ -359,6 +364,80 @@ protected APIResult update(Entity newEntity, String type, String entityName, Boo
...
+     *
+     * @param clusterName   Name of cluster
+     * @param colo colo
+     * @param skipDryRun Skip dry run during update if set to true
+     * @return APIResult
+     */
+    public APIResult updateClusterDependents(String clusterName, String colo, Boolean skipDryRun) {
+        checkColo(colo);
+        try {
+            verifySafemodeOperation(null, EntityUtil.ENTITY_OPERATION.UPDATE_CLUSTER_DEPENDENTS);
```
vrangan: Shouldn't the entityType be a valid entity type?

bvellanki: The operation is being performed on either Feed or Process entities. But the actual entity type should not matter for the operation, Hence sending null.

vrangan:
I am giving +1 with the following comments:   Let us fix the minor verifySafemode argument fix so that we are future proof
Let us add JIRA to add versions to other entities.   We initially wanted to have internal managed versions but having exposed them in cluster entity, we should do for others.

common/src/main/java/org/apache/falcon/entity/ClusterHelper.java
```diff
@@ -211,4 +219,55 @@ public static String getEmptyDir(Cluster cluster) {
...
+            for (Property prop : oldEntity.getProperties().getProperties()) {
+                oldProps.put(prop.getName(), prop.getValue());
+            }
+        }
+        if (newEntity.getProperties() != null) {
+            for (Property prop : newEntity.getProperties().getProperties()) {
+                newProps.put(prop.getName(), prop.getValue());
+            }
+        }
+
```
yzheng-hortonworks: Better to add a function to return properties of a cluster, as what you did for location and interface.

common/src/main/java/org/apache/falcon/entity/ClusterHelper.java
```diff
@@ -211,4 +219,55 @@ public static String getEmptyDir(Cluster cluster) {
...
+        if (oldProps.keySet().equals(newProps.keySet())) {
+            for (Map.Entry<String, String> keyVals : oldProps.entrySet()) {
+                if (!keyVals.getValue().equals(newProps.get(keyVals.getKey()))) {
+                    return false;
+                }
+            }
+            return true;
```
yzheng-hortonworks: Looks you are trying to compare if two maps are identical. You can use the existing function Map::equals (https://docs.oracle.com/javase/7/docs/api/java/util/Map.html#equals(java.lang.Object)).

bvellanki:
@vrangan @peeyushb @yzheng-hortonworks  : Latest patch includes all feedback you provided.

peeyushb:
Patch looks good. +1 .

yzheng-hortonworks:
+1

PR #128, opened 2016-05-06
FALCON-1935 Falcon fails to start with default startup.properties

PraveenAdlakha:
…nJPAService fails)


src/conf/statestore.properties
```diff
@@ -44,5 +42,22 @@
...
+*.falcon.statestore.pool.max.active.conn=10
+*.falcon.statestore.connection.properties=
+# Indicates the interval (in milliseconds) between eviction runs.
+*.falcon.statestore.validate.db.connection.eviction.interval=300000
+# The number of objects to examine during each run of the idle object evictor thread.
+*.falcon.statestore.validate.db.connection.eviction.num=10
+# Creates Falcon DB.
+# If set to true, it creates the DB schema if it does not exist. If the DB schema exists is a NOP.
+# If set to false, it does not create the DB schema. If the DB schema does not exist it fails start up.
+*.falcon.statestore.create.db.schema=true
```
pavankumar526: Make it false

src/conf/statestore.properties
```diff
@@ -16,14 +16,12 @@
 # limitations under the License.
 #
 
-*.domain=${falcon.app.type}
+*.domain=debug
```
sandeepSamudrala: why is this being changed?

PraveenAdlakha: @sandeepSamudrala  : The statestore.properties in schedule is set to debug mode by default. Thats why I changed it here.

bvellanki:
👍 thanks @PraveenAdlakha, I will commit this if there are no further comments.

yzheng-hortonworks:
+1. Thank you for addressing this, @PraveenAdlakha . Verified that this error fixed. However, I still got another error when it tries to create a directory for Derby DB (see below). Could you make the default setting work without extra Derby configuration (i.e. user can start Falcon without setting up Derby DB themselves)? One example is the embedded ActiveMQ server which we provide by default.

Caused by: ERROR XBM0H: Directory /usr/hdp/2.4.0.0-169/falcon-0.10-SNAPSHOT/target/test-data/data.db cannot be created.
	at org.apache.derby.iapi.error.StandardException.newException(Unknown Source)
	at org.apache.derby.iapi.error.StandardException.newException(Unknown Source)
	at org.apache.derby.impl.services.monitor.StorageFactoryService$10.run(Unknown Source)
	at java.security.AccessController.doPrivileged(Native Method)
	at org.apache.derby.impl.services.monitor.StorageFactoryService.createServiceRoot(Unknown Source)
	at org.apache.derby.impl.services.monitor.BaseMonitor.bootService(Unknown Source)
	at org.apache.derby.impl.services.monitor.BaseMonitor.createPersistentService(Unknown Source)
	at org.apache.derby.impl.services.monitor.FileMonitor.createPersistentService(Unknown Source)
	at org.apache.derby.iapi.services.monitor.Monitor.createPersistentService(Unknown Source)
	... 45 more

PraveenAdlakha:
@yzheng-hortonworks : I have figured out the issue why the exceptions are coming will remove FalconJPAService and FeedSLAMonitoritng from startup.properties as they are interdependent and we donot need both of them at the startup.Will add the required documentation how to enable the FeedSLAMonitring in a twiki.

yzheng-hortonworks:
Thank you, @PraveenAdlakha .

src/conf/statestore.properties
```diff
@@ -44,5 +44,7 @@
 ## Creates Falcon DB.
 ## If set to true, it creates the DB schema if it does not exist. If the DB schema exists is a NOP.
 ## If set to false, it does not create the DB schema. If the DB schema does not exist it fails start up.
-## In production environments it is recommended not to change this property,create schema using tool provided by falcon.
-#*.falcon.statestore.create.db.schema=false
+#*.falcon.statestore.create.db.schema=true
```
pavankumar526: Make it false as default and make it 1

pavankumar526:
+1 looks good.

PR #129, opened 2016-05-07
FALCON-1905 Extension Job Management: IT tests for REST APIs and CLIs

yzheng-hortonworks:


webapp/src/test/java/org/apache/falcon/resource/ExtensionManagerIT.java
```diff
@@ -0,0 +1,171 @@
...
+
+    @BeforeClass
+    public void prepare() throws Exception {
+        TestContext.prepare();
+        initExtensionStore(TestContext.class);
+    }
+
+    @AfterClass
+    public void tearDown() throws Exception {
+        TestContext.deleteEntitiesFromStore();
```
bvellanki: You might also want to call cleanup() to delete extensionStorePath during tearDown()

extensions/src/main/java/org/apache/falcon/extensions/store/ExtensionStore.java
```diff
@@ -56,7 +56,7 @@
     private static final String RESOURCES_DIR = "resources";
     private static final String LIBS_DIR = "libs";
 
-    private static final String EXTENSION_STORE_URI = "extension.store.uri";
+    public static final String EXTENSION_STORE_URI = "extension.store.uri";
```
bvellanki: Since this is a startup property, it could be cleaner to define it in StartupProperties. For example look at CONFIGSTORE_PROPERTY

bvellanki:
👍  @yzheng-hortonworks , will commit tonight.

webapp/src/test/resources/startup.properties
```diff
@@ -68,6 +68,7 @@ debug.retry.recorder.path=${user.dir}/target/retry
 debug.libext.feed.retention.paths=${falcon.libext}
 debug.libext.feed.replication.paths=${falcon.libext}
 debug.libext.process.paths=${falcon.libext}
+debug.extension.store.uri=jail://global:00/extension/store
```
sowmyaramesh: I see that config store is set to 
debug.config.store.uri=file://${user.dir}/target/store.

Instead of using jail FS can't we use file://${user.dir}/target/extension/store?

yzheng-hortonworks: @sowmyaramesh We can't use {user.dir}. It will run into an error because the initialization function in AbstractTestExtensionStore has a different {user.dir} value than the one where test falcon server is running. We need a consistent uri to initialize extension store. I choose to use fail FS because we also use this in test cluster definition (see webapp/src/test/resources/cluster-template.xml).

sowmyaramesh: +1

PR #132, opened 2016-05-10
FALCON-1940 Fix misc build issues in Falcon

bvellanki:


vrangan:
Thank @bvellanki .  I think It would be better to separate out pom fixes with CHANGELOG changes and comment changes for easier tracking

bvellanki:
@vrangan  : Fair point about changelog, I will remove it. I think the comment fix should stay in since the build fails with test-patch profile. Please let me know if this is ok and I will make the changes.

vrangan:
Agreed.  It was not clear why the second fix was there.   Thanks for pointing out.  +1 after removing changelog

PR #136, opened 2016-05-11
FALCON-1953 Build fails when profiles hivedr and test-patch are used together

bvellanki:


vramachan:
+1.
Looks good.

vrangan:
+1 -  Sorry missed running with the profile earlier

PR #138, opened 2016-05-12
FALCON-1919 Provide user the option to store sensitive information with Hadoop credential provider

yzheng-hortonworks:
Tested alias properties being resolved correctly with Hadoop credential provider.

### DIFF
--- a/common/src/main/java/org/apache/falcon/util/StartupProperties.java
+++ b/common/src/main/java/org/apache/falcon/util/StartupProperties.java
@@ -20,8 +20,6 @@ package org.apache.falcon.util;
 
 import org.apache.falcon.FalconException;
 import org.apache.falcon.hadoop.HadoopClientFactory;
-import org.apache.falcon.security.CredentialProviderHelper;
-import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.fs.FileSystem;
 import org.apache.hadoop.fs.Path;
 import org.slf4j.Logger;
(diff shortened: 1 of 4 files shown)


common/src/main/java/org/apache/falcon/util/StartupProperties.java
```diff
@@ -37,6 +39,9 @@
     public static final String SAFEMODE_PROPERTY = "falcon.safeMode";
     private static final String SAFEMODE_FILE = ".safemode";
     private static final String CONFIGSTORE_PROPERTY = "config.store.uri";
+    private static final String CREDENTIAL_PROVIDER_PROPERTY = "hadoop.security.credential.provider.path";
+    private static final String ALIAS_PROPERTY_PREFIX = "hadoop.security.alias.";
```
vrangan: Instead of saying hadoop.security.alias we should just say "credential.provider.alias.for" (which allows us to decouple from implementation specifics in future in terms of names).   The for at end makes it easy to understand the purpose

vrangan:
Thanks for doing this.   Add a unit testcase for the credential providing alias resolution - Please see TestPassingSecurePassword#testCredentialProviderLoader and TestPassingSecurePassword#testPasswordAliasOption in Sqoop

In fact, we no longer need the reflection mechanisms as Falcon is now using Hadoop 2.6+

vrangan:
Also, let us add a documentation for this

yzheng-hortonworks:
Thanks for your review and reminder, @vrangan . I've created a JIRA to document the usage: https://issues.apache.org/jira/browse/FALCON-1957

yzheng-hortonworks:
@vrangan  Added unit test. Thanks for the pointers. Also created a jira to remove unneeded reflection mechanisms: https://issues.apache.org/jira/browse/FALCON-1959

vrangan:
+1  - The build failures are unrelated.

bvellanki:
+1 thanks for adding tests as well. Will commit shortly

PR #141, opened 2016-05-13
FALCON-1956 Graphite Plugin for monitoring

PraveenAdlakha:


vrangan:
+1 for Graphite support.

As a follow up, We probably need to see how we can support different metrics providers.   I am primarily thinking about AMS, as it is mostly available for most of the Falcon users (as it is part of Ambari) and integrated with rest of the Hadoop metrics and now with Grafana dashboards

prism/src/main/java/org/apache/falcon/plugin/GraphiteNotificationPlugin.java
```diff
@@ -0,0 +1,79 @@
...
+import org.joda.time.DateTime;
+import org.joda.time.Seconds;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+
+/**
+ * Graphite Notification Plugin.
+ */
+public class GraphiteNotificationPlugin implements MonitoringPlugin {
```
sriksun: Please do consider Adding an AlertingPlugin additionally to capture permanent failures as well.

sriksun:
:+1:

prism/src/main/java/org/apache/falcon/plugin/GraphiteNotificationPlugin.java
```diff
@@ -0,0 +1,79 @@
...
+            String prefix = StartupProperties.get().getProperty("falcon.graphite.prefix");
+            if (entityType.equals(EntityType.PROCESS.name())) {
+                Entity entity = ConfigurationStore.get().get(EntityType.PROCESS, entityName);
+                Process process = (Process) entity;
+                String pipeline =  StringUtils.isNotBlank(process.getPipelines()) ? process.getPipelines() : "default";
+
+
+                if ((message.getAction().equals("wf-instance-succeeded"))) {
+                    Long timeTaken =  message.getExecutionTime() / 1000000000;
+                    String metricsName = prefix + pipeline + ".GENERATE." + entityName + ".processing_time";
```
pavankumar526: All time metrics are in secs ? We should be consistent in this

prism/src/main/java/org/apache/falcon/plugin/GraphiteNotificationPlugin.java
```diff
@@ -0,0 +1,79 @@
...
+                    metricNotificationService.publish(metricsName, timeTaken);
+
+                    DateTime nominalTime = new DateTime(message.getDimensions().get("nominal-time"));
+                    DateTime startTime = new DateTime(message.getDimensions().get("start-time"));
+                    metricsName = prefix + pipeline + ".GENERATE." + entityName + ".start_delay";
```
pavankumar526: Why GENERATE this is added ?

pavankumar526:
Looks good otherwise

prism/src/main/java/org/apache/falcon/plugin/GraphiteNotificationPlugin.java
pavankumar526: Add cluster name as we discussed.

pavankumar526:
+1 looks good

PR #143, opened 2016-05-16
FALCON-1858: Support HBase as a storage backend for Falcon Titan graphDB

vrangan:
This patch provides the following features
1.   Ability to build Falcon with storage backend plugins without the artifacts (hbase and bdb).
2.   Ability to package a standalone HBase if desired
3.  Backport of Titan 1.0 Hbase 1.0 storage backend into Titan 0.5.4 (shaded classes - mostly similar to Apache Atlas usage of the same) (most of the addons/titan codebase)
4.  Default build profile is with bdb backend with bdb artifacts also included in the build.

Tested the HBase backend with 1.1.5 and 1.2.1 Hbase environments.   Ran test-patch successfully with enabling/disabling of bdb db artifacts and also with all potential addons enabled

### DIFF
--- a/src/conf/startup.properties
+++ b/src/conf/startup.properties
@@ -170,6 +170,9 @@ prism.configstore.listeners=org.apache.falcon.entity.v0.EntityGraph,\
 *.falcon.graph.blueprints.graph=com.thinkaurelius.titan.core.TitanFactory
 
 # Graph Storage
+# IMPORTANT:   Please enable one of hbase or berkeleydb backends are enabled
+#  after the backend requirements are provisioned as needed.
+
 # Enable the following for Berkeley DB.  Make sure je-5.0.73.jar is
 # downloaded and available under Falcon webapp directory or under falcon
 # server classpath.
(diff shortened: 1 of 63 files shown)


src/conf/startup.properties
```diff
@@ -168,9 +168,20 @@ prism.configstore.listeners=org.apache.falcon.entity.v0.EntityGraph,\
...
 # Graph Storage
-*.falcon.graph.storage.directory=/${falcon.home}/data/graphdb
-*.falcon.graph.storage.backend=berkeleyje
+# Berkeley DB
+#*.falcon.graph.storage.directory=/${falcon.home}/data/graphdb
+#*.falcon.graph.storage.backend=berkeleyje
+
+
+# Enabled the following for HBase
+*.falcon.graph.storage.backend=hbase
```
peeyushb: Here hbase as storage backend is enabled. I am of the opinion that Berkley should be enabled by default.

peeyushb:
@vrangan  I have done review of the pull request. On revieweing I found that docs, how to build the Falcon with hbase or with berkley profile to use it as a storage backend, is not available.

src/conf/startup.properties
vrangan: We should not have either of them enabled by default in the release versions and have users download the necessary jars.  Have updated the docs

peeyushb:
When I have used the patch from this pull request , Falcon build is proper. But when I am trying to start the Falcon server it is not starting on standalone machine.

vrangan:
Discussed with @peeyushb  - we have one hbase property enabled by default in startup.properties.   That requires standalone hbase with default properties (as the other hbase related properties are commented out).  So, as per comment in the properties we will comment all the graphdb related properties and users will have to uncomment one or the other backend.

Note that hadoop installer can make this automatic without users needing to do it.

vrangan:
Made a change after discussion with @peeyushb.   Will follow up with a documentation JIRA to augment how to standup a hbase standalone mode cluster

peeyushb:
Patch looks good. +1

bvellanki:
+1

peeyushb:
I will commit this patch soon, if no further comments.

PR #144, opened 2016-05-16
FALCON-1961 Should return error if an extension job doesn't exist for delete/suspend/resume/schedule operations

yzheng-hortonworks:


prism/src/main/java/org/apache/falcon/resource/extensions/ExtensionManager.java
```diff
@@ -241,6 +256,11 @@ public APIResult delete(@PathParam("job-name") String jobName,
                             @DefaultValue("") @QueryParam("doAs") String doAsUser) {
         try {
             List<Entity> entities = getEntityList("", "", "", TAG_PREFIX_EXTENSION_JOB + jobName, "", doAsUser);
+            if (entities.size() == 0) {
+                // return failure if the extension job doesn't exist
+                return new APIResult(APIResult.Status.FAILED, "Extension job " + jobName + " doesn't exist.");
```
bvellanki: Instead of returning APIResult.Status.FAILED, you should return 

APIResult(APIResult.Status.SUCCEEDED, entity + "(" + type + ") doesn't exist. Nothing to do");

Please refer to AbstractEntityManager

yzheng-hortonworks: Changed to return SUCCEEDED status

bvellanki:
+1

sowmyaramesh:
@yzheng-hortonworks : Even for getInstances and getExtensionJobs if getEntityList returns empty list then it should return  and not continue further processing.

yzheng-hortonworks:
@sowmyaramesh If there is no job for list and instances operations, it will return 0 results with current code. So the current results for these two APIs you mentioned are already expected.

yzheng-hortonworks:
@sowmyaramesh Added "return immediately" in list and instances.

sowmyaramesh:
+1, Thanks for your contribution. Will commit.

PR #145, opened 2016-05-16
FALCON-1943 Extension API/CLI fails when authorization is enabled

sowmyaramesh:


vrangan:
+1 - We should allow extension as a top level resource

yzheng-hortonworks:
+1

PR #146, opened 2016-05-16
FALCON-1964 Should delete temporary JKS file after IT tests for credential provider alias

yzheng-hortonworks:


common/src/test/java/org/apache/falcon/util/ApplicationPropertiesTest.java
```diff
@@ -67,6 +67,12 @@ public void testResolveAlias() throws Exception {
         properties.resolveAlias();
         Assert.assertEquals(properties.getProperty(PROPERTY_1), PASSWORD_1);
         Assert.assertEquals(properties.getProperty(PROPERTY_2), PASSWORD_2);
+
+        // delete temporary jks files
```
bvellanki: Can you please add it as @AfterClass to ensure these files are deleted even if there are test failures

yzheng-hortonworks: Thanks for pointing this out, @bvellanki . Added to AfterClass.

bvellanki:
+1

common/src/test/java/org/apache/falcon/util/ApplicationPropertiesTest.java
```diff
@@ -39,6 +40,16 @@
     private static final String PROPERTY_2 = "property-key-2";
     private static final String JKS_FILE_NAME = "credentials.jks";
 
+    @AfterClass
+    public void tearDown() throws Exception {
+        // delete temporary jks files
+        File credDir = new File(".");
```
sowmyaramesh: Can we define credDir as public static final String CRED_DIR so that it can be reused in tearDown and testResolveAlias? Thanks!

yzheng-hortonworks: Addressed.

sowmyaramesh:
+1, will commit.

PR #147, opened 2016-05-17
FALCON-1963 Falcon CLI should provide detailed hints if the user's command is invalid

yzheng-hortonworks:


bvellanki:
+1, the jenkins failure is unrelated to the patch. Will commit shortly if there are no further comments.

PR #151, opened 2016-05-19
FALCON-1974 Cluster update : Allow superuser to update bundle/coord of dependent entities

bvellanki:


oozie/src/main/java/org/apache/falcon/workflow/engine/OozieWorkflowEngine.java
```diff
@@ -1434,34 +1434,40 @@ private Frequency getDelay(Feed entity, CoordinatorJob coord) {
...
+        try {
+            //Validate that new entity can be scheduled
+            dryRunForUpdate(cluster, newEntity, effectiveTime, skipDryRun);
+
+            boolean suspended = BUNDLE_SUSPENDED_STATUS.contains(oldBundle.getStatus());
+
+            //Set end times for old coords
+            updateCoords(clusterName, oldBundle, EntityUtil.getParallel(oldEntity), effectiveTime, newEntity);
+            //schedule new entity
+            String newJobId = scheduleForUpdate(newEntity, cluster, effectiveTime, user);
```
yzheng-hortonworks: The function scheduleForUpdate also performs user switch (current user -> user -> current user). However, since you already switch the user before calling scheduleForUpdate, the switch is no longer needed in scheduleForUpdate. Better to remove the "user" argument and user switch in scheduleForUpdate, as this function is only being used in updateInternal.

oozie/src/main/java/org/apache/falcon/workflow/engine/OozieWorkflowEngine.java
```diff
@@ -1434,34 +1434,40 @@ private Frequency getDelay(Feed entity, CoordinatorJob coord) {
...
-        //Sometimes updateCoords() resumes the suspended coords. So, if already suspended, resume now
-        //Also suspend new bundle
-        if (suspended) {
-            doBundleAction(newEntity, BundleAction.SUSPEND, cluster.getName());
+            //Sometimes updateCoords() resumes the suspended coords. So, if already suspended, resume now
+            //Also suspend new bundle
+            if (suspended) {
+                doBundleAction(newEntity, BundleAction.SUSPEND, cluster.getName());
+            }
+            switchUser(currentUser);
```
yzheng-hortonworks: You don't need to call "switchUser" here. You already add it to the "finally" block :)

oozie/src/main/java/org/apache/falcon/workflow/engine/OozieWorkflowEngine.java
bvellanki: Good catch @yzheng-hortonworks , I will make the necessary update.

oozie/src/main/java/org/apache/falcon/workflow/engine/OozieWorkflowEngine.java
bvellanki: Yes, you are right.

yzheng-hortonworks:
+1

PR #152, opened 2016-05-19
FALCON-1972 Handling cases when Extension service or extension.store.uri is not set in startup proeprties

sowmyaramesh:


bvellanki:
+1

PR #157, opened 2016-05-22
Falcon 1949 : Update Data source Entity

vramachan:
+ Updates data source entity in the config store
+ Rewrites the corresponding IMPORT and EXPORT bundles with end and start dates for old and new bundles.
+ UT and IT
+ Manual end-to-end testing.


vramachan:
org.apache.falcon.metadata.MetadataMappingServiceTest.testOnClusterEntityChange seems to be flaky. It passes on my local builds, but fails on the Jenkins.

common/src/main/java/org/apache/falcon/entity/DatasourceHelper.java
```diff
@@ -147,17 +151,140 @@ public static String buildJceksProviderPath(URI credURI) {
...
+        }
+        switch (cred.getType()) {
+        case PASSWORD_TEXT:
+            if (StringUtils.isBlank(cred.getUserName()) || StringUtils.isBlank(cred.getPasswordText())) {
+                throw new FalconException(String.format("Credential type '%s' missing tags '%s' or '%s'",
+                    cred.getType().value(), "userName", "passwordText"));
+            }
+            break;
+        case PASSWORD_FILE:
+            if (StringUtils.isBlank(cred.getUserName()) || StringUtils.isBlank(cred.getPasswordFile())) {
```
bvellanki: Should we also check if the file exists and readable?

common/src/main/java/org/apache/falcon/entity/DatasourceHelper.java
```diff
@@ -147,17 +151,140 @@ public static String buildJceksProviderPath(URI credURI) {
...
+        if ((StringUtils.isBlank(ep1)) && (StringUtils.isBlank(ep2))) {
+            return true;
+        }
+        if ((StringUtils.isBlank(ep1)) && (StringUtils.isBlank(ep2))) {
+            return false;
+        }
+        return ep1.equals(ep2);
+
+    }
     /**
```
bvellanki: Minor nit : Some space between methods would be more readable

common/src/main/java/org/apache/falcon/entity/parser/DatasourceEntityParser.java
```diff
@@ -51,6 +51,10 @@ public DatasourceEntityParser() {
     public void validate(Datasource db) throws FalconException {
         try {
             ClassLoader hdfsClassLoader = HdfsClassLoader.load(db.getName(), db.getDriver().getJars());
+            DatasourceHelper.validateCredential(DatasourceHelper.getCredential(db));
```
bvellanki: It might be more readable to make it DatasourceHelper.validateCredentials(...) and internally validate all types.

common/src/main/java/org/apache/falcon/metadata/EntityRelationshipGraphBuilder.java
```diff
@@ -80,7 +80,7 @@ public void addClusterEntity(Cluster clusterEntity) {
         Vertex clusterVertex = addVertex(clusterEntity.getName(), RelationshipType.CLUSTER_ENTITY);
 
         addUserRelation(clusterVertex);
-        addColoRelation(clusterEntity.getColo(), clusterVertex);
+        addColoRelation(clusterEntity.getColo(), clusterVertex, RelationshipLabel.CLUSTER_COLO);
```
bvellanki: addColoRelation already sets correct relationship label. You wont have to pass it as arg.

        addEdge(fromVertex, coloVertex, RelationshipLabel.CLUSTER_COLO.getName());

common/src/main/java/org/apache/falcon/metadata/EntityRelationshipGraphBuilder.java
```diff
@@ -111,7 +111,7 @@ public void addDatasourceEntity(Datasource dsEntity) {
         Vertex dsVertex = addVertex(dsEntity.getName(), RelationshipType.DATASOURCE_ENTITY);
 
         addUserRelation(dsVertex);
-        addColoRelation(dsEntity.getColo(), dsVertex);
+        addColoRelation(dsEntity.getColo(), dsVertex, RelationshipLabel.DATASOURCE_COLO);
```
bvellanki: same as above

common/src/main/java/org/apache/falcon/update/UpdateHelper.java
```diff
@@ -166,4 +168,36 @@ public static boolean isClusterEntityUpdated(final org.apache.falcon.entity.v0.c
 
         return false;
     }
+
+    public static boolean isDatasourceEntityUpdated(final Datasource oldEntity, final Datasource newEntity)
+        throws FalconException {
+        // ignore changes : colo, acl, description, tags
```
bvellanki: colo update should not be ignored.

prism/src/main/java/org/apache/falcon/resource/AbstractEntityManager.java
```diff
@@ -354,6 +357,12 @@ protected APIResult update(Entity newEntity, String type, String entityName, Boo
             }
 
             configStore.update(entityType, newEntity);
+            // for Data Sources, check always if dependant feeds are already upgraded and upgrade accordingly
+            if (entityType.equals(EntityType.DATASOURCE)) {
```
bvellanki: Minor suggestion feel free to ignore : We have different tasks to be performed for different entityType. May be having a switch on entityType and moving corresponding update commands to the right case will make it more readable.

prism/src/main/java/org/apache/falcon/resource/AbstractEntityManager.java
```diff
@@ -354,6 +357,12 @@ protected APIResult update(Entity newEntity, String type, String entityName, Boo
...
+                ConfigurationStore.get().cleanupUpdateInit();
+                releaseEntityLocks(entityName, tokenList);
+                updateDatasourceDependents(entityName, skipDryRun);
```
bvellanki: If it is a single-step update, the dependents are updated before the configStore is updated. Else, if the updateDatasourceDependents fails midway, there is no way to recover.

prism/src/main/java/org/apache/falcon/resource/AbstractEntityManager.java
```diff
@@ -365,6 +374,81 @@ protected APIResult update(Entity newEntity, String type, String entityName, Boo
...
+                            if (updatedFeed) {
+                                // rewrite the dependent feed and update it on the store
+                                result.append(getWorkflowEngine(entity).update(entity, newFeed,
+                                        feedCluster.getName(), skipDryRun));
+                                updateEntityInConfigStore(entity, newFeed);
+                            }
+                        }
+                    }
+                }
+            }
```
bvellanki: Why check if type is FEED? isnt it always feed? Should we throw an error if it is not FEED?

common/src/main/java/org/apache/falcon/update/UpdateHelper.java
vramachan: data source does not have any colo context and impact like cluster.

prism/src/main/java/org/apache/falcon/resource/AbstractEntityManager.java
vramachan: Datasource entity is updated in the config store (line:359) and then all the dependent feeds are regenerated and stored  in configstore one by one.

common/src/main/java/org/apache/falcon/entity/parser/DatasourceEntityParser.java
vramachan: not sure if I understand it correctly. In that case, we would need a method like validateCredentialHelper(Credential)? this is because the way the xsd provides credentials for each interface.

common/src/main/java/org/apache/falcon/entity/DatasourceHelper.java
vramachan: validateInterface() method actually checks for the file and reads the password to connect and validate the db connection. Yes, it is done.

prism/src/main/java/org/apache/falcon/resource/AbstractEntityManager.java
vramachan: will assert that.

prism/src/main/java/org/apache/falcon/resource/AbstractEntityManager.java
vramachan: rearranged.

vramachan:
Updated patch available.

common/src/main/java/org/apache/falcon/update/UpdateHelper.java
bvellanki: Thank you, you are right about it.

common/src/main/java/org/apache/falcon/entity/DatasourceHelper.java
```diff
@@ -148,16 +152,168 @@ public static String buildJceksProviderPath(URI credURI) {
...
+                throw new FalconException(String.format("Credential type '%s' missing tags '%s' or '%s' or %s'",
+                    cred.getType().value(), "userName", "alias", "providerPath"));
+            }
+            break;
+        default:
+            throw new FalconException(String.format("Unknown Credential type '%s'", cred.getType().value()));
+        }
+    }
+
+    private static boolean isSameString(String ep1, String ep2) {
```
yzheng-hortonworks: You can use existing function StringUtils::equals. The null case is already handled there.

common/src/main/java/org/apache/falcon/entity/DatasourceHelper.java
```diff
@@ -148,16 +152,168 @@ public static String buildJceksProviderPath(URI credURI) {
...
      * Return the Interface endpoint for the interface type specified in the argument.
      *
      * @param db
      * @param type - can be read-only or write
      * @return
      */
-    private static String getInterface(Datasource db, Interfacetype type) {
+    private static String getInterfaceEndpoint(Datasource db, Interfacetype type) {
+        if (getInterface(db, type) != null) {
+            return getInterface(db, type).getEndpoint();
```
yzheng-hortonworks: Better to record the interface getInterface(db, type) so it doesn't need to be computed twice.

vramachan: Java/JVM should handle that optimization.

yzheng-hortonworks:
+1

common/src/main/java/org/apache/falcon/entity/v0/EntityGraph.java
```diff
@@ -200,6 +200,15 @@ public void onReload(Entity entity) throws FalconException {
                 feedEdges.add(dbNode);
                 dbEdges.add(feedNode);
             }
+            if (FeedHelper.isExportEnabled(cluster)) {
```
bvellanki: https://github.com/apache/falcon/pull/165/files seems to have same changes for EntityGraph.java, can you make sure only one of the two pull requests have this change?

bvellanki:
EntityGraph.java has code that also went into pull request 165. +1 once you remove the duplicate code.

bvellanki:
+1 Will commit shortly if there are no further comments. This is going in to master branch and not into 0.10

PR #158, opened 2016-05-24
FALCON-1978 : Fix flaky unite test MetadataMappingServicesTests

vramachan:
1. Ordered the tests so that parallel execution will not cause assert failures
2. There was an issue found in EntityGraphRelationshipBuilder that was causing MetadataMappingServicesTests.testOnClusterEntityChange(), but it seems that has been fixed in the trunk.


bvellanki:
Since the counts of edge and vertices are made based on following code in each method,  running tests in parallel is prone to flakiness. 

long beforeVerticesCount = getVerticesCount(service.getGraph());
long beforeEdgesCount = getEdgesCount(service.getGraph());

The fix makes sense. 👍

bvellanki:
MetadataMappingServicesTest has passed in jenkins. The failures are unrelated to this fix. Will commit shortly.

PR #160, opened 2016-05-25
FALCON-1980 : Change input and output argument order for Spark process workflow

peeyushb:
Details to perform this change has been mentioned in the FALCON-1980.

### DIFF
--- a/oozie/src/main/java/org/apache/falcon/oozie/process/SparkProcessWorkflowBuilder.java
+++ b/oozie/src/main/java/org/apache/falcon/oozie/process/SparkProcessWorkflowBuilder.java
@@ -89,7 +89,7 @@ public class SparkProcessWorkflowBuilder extends ProcessExecutionWorkflowBuilder
             argList.addAll(sparkArgs);
         }
 
-        //Adding output first so that final order must have input first and then output followed by user's arguments.
+        //Adding output first so that final order must have input and then output followed by user's arguments.
         addOutputFeedsAsArgument(argList, cluster);
         addInputFeedsAsArgument(argList, cluster);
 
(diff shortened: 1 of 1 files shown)


oozie/src/main/java/org/apache/falcon/oozie/process/SparkProcessWorkflowBuilder.java
```diff
@@ -163,13 +164,16 @@ private void addOutputFeedsAsArgument(List<String> argList, Cluster cluster) thr
             return;
         }
 
-        for(Output output : entity.getOutputs().getOutputs()) {
+        int numOutputFeed = entity.getOutputs().getOutputs().size();
```
pavankumar526: I can understand you want to put input first and then output , but with in output or input ordering should be maintained right ?

peeyushb: @pavankumar526  Being I want input first and then output, I am doing reverse insertion. Inserting output first at 0 index and then input at 0 index in the arg List. So with this, final List will appear in order as input, output and user's argument. As number of arguments are not much, so there will not be much impact in List insertion at 0 index.

pavankumar526:
+1 looks good

peeyushb:
Thanks @pavankumar526 for the review. If no further comments, I will commit this patch soon.

oozie/src/main/java/org/apache/falcon/oozie/process/SparkProcessWorkflowBuilder.java
vrangan: Can you add a comment on this (that we are adding to the 0th index and getting the args shifted as arguments are added to get the desired effect).

peeyushb: Comment added.

PR #161, opened 2016-05-25
FALCON-1981 : Remove runtime jar dependencies from pom files

vramachan:
+ Remove jar dependency from pom files that are not required during runtime, but provided by Hadoop
+ End to End test to make sure few flows work


yzheng-hortonworks:
+1

vrangan:
+1

PR #162, opened 2016-05-26
FALCON-1985: Disable FeedExportIT and FeedImportIT temporarily

vramachan:
These tests started to fail since may due to some infrastructure changes on the jenkins servers. debugging that will require access to those servers. until then disabling these tests to unblock 0.10 release branch creation.


bvellanki:
Build failed, you might have to set enable to false for each individual test.

bvellanki:
+1 checks passed. Will commit shortly

PR #163, opened 2016-05-26
FALCON-1983 Upgrade jackson core and databind versions to fix dependency incompatibility with higher-version Hive

yzheng-hortonworks:
Tested it works and generates consistent jackson jars with Hive 1.2.1x.


bvellanki:
+1 thanks Ying

vrangan:
I initially thought excluding jackson* from hive would be enough, but it looks like we are moving to jackson 2 which is good in a way. 

+1

PR #165, opened 2016-05-27
FALCON-1859 : Feed export instances are not added to Graph DB

vramachan:


bvellanki:
+1 looks good to me. Will commit to master if there are no further comments.

@vramachan : Why should this Jira need to go into 0.10?

bvellanki:
Spoke with @vramachan , it is a bug and not a new feature. Will commit to 0.10 and master.

PR #166, opened 2016-06-01
FALCON-1984 Provide proper hint and documentation if required titan storage backend is not configured in startup.properties

yzheng-hortonworks:
Added validation of required configuration properties for graph db storage backend.
Added error hints on the missing property in Falcon server log.
Reorganized the order of the configuration properties in startup.properties so relevant properties for each backend are grouped in consecutive lines.

### DIFF
--- a/common/src/main/java/org/apache/falcon/metadata/MetadataMappingService.java
+++ b/common/src/main/java/org/apache/falcon/metadata/MetadataMappingService.java
@@ -35,6 +35,7 @@ import com.tinkerpop.blueprints.util.TransactionWork;
 import org.apache.commons.configuration.BaseConfiguration;
 import org.apache.commons.configuration.Configuration;
 import org.apache.falcon.FalconException;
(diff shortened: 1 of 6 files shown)


common/src/main/java/org/apache/falcon/metadata/MetadataMappingService.java
```diff
@@ -118,11 +141,56 @@ public void init() throws FalconException {
 
     protected Graph initializeGraphDB() {
         LOG.info("Initializing graph db");
-
         Configuration graphConfig = getConfiguration();
+        validateConfiguration(graphConfig);
         return GraphFactory.open(graphConfig);
     }
 
+    private void validateConfiguration(Configuration graphConfig) {
```
bvellanki: The error messages seem similar, minor nit.

bvellanki:
+1

vrangan:
+1

PR #167, opened 2016-06-01
FALCON-2007 : Hive DR Replication failing with - Can not create a Path from a null string

peeyushb:
Following is the pull request that will check the existence of the file oozie.action.conf.xml. If file is not there, then it will not get into the code block that adds it to the configuration object, thus to avoid the exception.


vrangan:
+1

peeyushb:
If no further comments on this pull request, I will commit this patch soon so as to avoid further failures in secure environment.

PR #168, opened 2016-06-03
FALCON-1965 Update ActiveMQ version to 5.13.3 to avoid Falcon start error after rolling upgrade

yzheng-hortonworks:
Upgraded ActiveMQ version to the latest 5.13.3. Tested Falcon server upgrade with ActiveMQ history works with 0.6, 0.7 and trunk.

Split the fix for two test failures due to upgrade to https://github.com/apache/falcon/pull/169, per @vrangan 's suggestion.


bvellanki:
Agree with @yzheng-hortonworks , +1. Will commit if there are no objections to changing the tests.

vrangan:
+1.   I would say split it into two jiras one for upgrade and one for the tests fix so that it is more easily trackable.

yzheng-hortonworks:
Thank you for your suggestion, @vrangan . I split the UT fix to a separate pull request for ease of tracking. https://github.com/apache/falcon/pull/169

pallavi-rao:
+1

PR #169, opened 2016-06-03
FALCON-2010 Fix UT errors due to ActiveMQ upgrade

yzheng-hortonworks:
Fixed two test failures due to the upgrade:

1. JMSMessageConsumerTest
After upgrade, some messages are consumed faster and so the initial expectation on no message being dequeued right after the message sending doesn't hold. Also confirmed with @bvellanki that this kind of tests are flaky and have caused intermittent test errors before. Therefore, removed getTotalDequeueCount assertion. In addition, the expected message consumptions are tested in existing code "Mockito.verify(jobEndService,...".

2. FalconPostProcessingTest
After upgrade, Falcon will run into an error if it tries to delete messages when ActiveMQ service is stopped. Therefore, need to call start before deleting the message during teardown.

### DIFF
--- a/messaging/src/test/java/org/apache/falcon/messaging/JMSMessageConsumerTest.java
+++ b/messaging/src/test/java/org/apache/falcon/messaging/JMSMessageConsumerTest.java
@@ -245,15 +245,10 @@ public class JMSMessageConsumerTest {
             sendMessages(TOPIC_NAME, WorkflowExecutionContext.Type.POST_PROCESSING);
 
             final BrokerView adminView = broker.getAdminView();
-
-            Assert.assertEquals(adminView.getTotalDequeueCount(), 0);
-//            Assert.assertEquals(adminView.getTotalEnqueueCount(), 10);
             Assert.assertEquals(adminView.getTotalConsumerCount(), 2);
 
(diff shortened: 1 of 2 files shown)


messaging/src/test/java/org/apache/falcon/messaging/JMSMessageConsumerTest.java
```diff
@@ -309,7 +298,6 @@ public void testJMSMessagesFromOozieForNonFalconWF() throws Exception {
         sendMessages(TOPIC_NAME, WorkflowExecutionContext.Type.WORKFLOW_JOB, false /* isFalconWF */);
 
         final BrokerView adminView = broker.getAdminView();
-
         Assert.assertEquals(adminView.getTotalDequeueCount(), 0);
```
pallavi-rao: May be remove it here too to be consistent and also so the test doesn't fail later.

pallavi-rao:
+1 for the patch other than the minor comment.

messaging/src/test/java/org/apache/falcon/messaging/JMSMessageConsumerTest.java
yzheng-hortonworks: Removed and keep only consumer counts. Thanks for your review, @pallavi-rao !

bvellanki:
+1

PR #176, opened 2016-06-08
FALCON-2014 Issue with submit idempotency in distributed mode.

pallavi-rao:


sandeepSamudrala:
The same needs to be done for delete and update too.

sandeepSamudrala:
Jenkins check has failed for the same for testUpdateAndTouch and testDelete.

pallavi-rao:
The check makes no sense for delete. For update, the current stand is that we just overwrite old def with whatever the user supplies. So, not applicable for update either.

The tests are failing because they have duplicate submit statements. The fix just uncovered that mistake.

sandeepSamudrala:
Makes sense.

pallavi-rao:
I'm closing this pull request, as addressing the problem is not as simple as I initially thought. Will create a new pull request later.

PR #181, opened 2016-06-09
FALCON-2008 Add documentation for Graphite Notification Plugin

PraveenAdlakha:


pallavi-rao:
+1. Will merge if no further comments

PR #182, opened 2016-06-11
FALCON-2023 : Feed eviction fails when feed locations stats and meta …

vramachan:
…does not have time pattern

### DIFF
--- a/retention/src/test/java/org/apache/falcon/retention/FeedEvictorTest.java
+++ b/retention/src/test/java/org/apache/falcon/retention/FeedEvictorTest.java
@@ -462,6 +462,42 @@ public class FeedEvictorTest {
         }
     }
 
+    @Test
+    public void testEvictionStatsMetaWithNoPattern() throws Exception {
+        try {
+            Configuration conf = cluster.getConf();
+            FileSystem fs = FileSystem.get(conf);
+            fs.delete(new Path("/"), true);
+            stream.clear();
+
+            Pair<List<String>, List<String>> pair = createTestData("/data");
+            createDir("/stats");
+            createDir("/meta");
+            createTestData("/tmp");
+            final String storageUrl = cluster.getConf().get(HadoopClientFactory.FS_DEFAULT_NAME_KEY);
+            FeedEvictor.main(new String[] {
+                    "-feedBasePath", getFeedBasePath(LocationType.DATA, storageUrl)
+                    + "#" + getStatsOrMetaPath(LocationType.STATS, storageUrl)
+                    + "#" + getStatsOrMetaPath(LocationType.META, storageUrl)
+                    + "#" + getFeedBasePath(LocationType.TMP, storageUrl),
+                    "-retentionType", "instance",
+                    "-retentionLimit", "months(5)",
+                    "-timeZone", "UTC",
+                    "-frequency", "hourly",
+                    "-logFile", conf.get(HadoopClientFactory.FS_DEFAULT_NAME_KEY)
+                    + "/falcon/staging/feed/2012-01-01-04-00", "-falconFeedStorageType",
+                    Storage.TYPE.FILESYSTEM.name(),
+            });
+
+            // should not throw exception
+            // stats and meta dir should not be deleted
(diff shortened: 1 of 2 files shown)


sandeepSamudrala:
Indentation checks seems to fail. Please fix them
Changes look good to me.

vramachan:
@sandeepSamudrala @bvellanki please review.

bvellanki:
@vramachan : There should be an error when data does not have time-pattern. It should handle meta and stats slightly differently from data. This fix seems to let eviction pass even if location data does not have time pattern. Please correct me if I am wrong.

retention/src/test/java/org/apache/falcon/retention/FeedEvictorTest.java
```diff
@@ -462,6 +462,43 @@ public void testFeedBasePathExists() throws Exception {
         }
     }
 
+    @Test
+    public void testEvictionStatsMetaWithNoPattern() throws Exception {
+        try {
+            Configuration conf = cluster.getConf();
+            FileSystem fs = FileSystem.get(conf);
+            fs.delete(new Path("/"), true);
```
vramachan: @bvellanki The feed definition (FeedEntityParser) currently allows data location without patterns. (see below the example). If we need to enforce data location to have patterns, it should be done at the time of submission. I'm reluctant to do that now since it could have backward compatibility issues and should be done via a separate discuss/JIRA.

<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
<feed name="testfeed-data-stats-meta-nopattern" description="Raw customer email feed" xmlns="uri:falcon:feed:0.1">
    <tags>externalSystem=USWestEmailServers</tags>
    <groups>churnAnalysisDataPipeline</groups>
    <frequency>hours(1)</frequency>
    <timezone>UTC</timezone>
    <late-arrival cut-off="hours(1)"/>
    <clusters>
        <cluster name="primaryCluster" type="source">
            <validity start="2016-04-30T01:00Z" end="2017-10-30T10:00Z"/>
            <retention limit="days(1)" action="delete"/>
        </cluster>
    </clusters>
    <locations>
        <location type="data"  path="/tmp/falcon-regression/testfeed-stats-meta-nopattern/data/"/>
        <location type="stats" path="/tmp/stats"/>
        <location type="meta"  path="/tmp/meta"/>
    </locations>
    <ACL owner="ambari-qa" group="users" permission="0x755"/>
    <schema location="/none" provider="/none"/>
</feed>

vramachan: @bvellanki there are other tests doing this and it seems to work. If there is any race condition, all the tests in this class need to be sequenced one after the other. I'm not planning to address it now.

   Configuration conf = cluster.getConf();
+            FileSystem fs = FileSystem.get(conf);
+            fs.delete(new Path("/"), true);

bvellanki: Can you please make that change now? It should not take long and it would save us from potential pain in the future.

vramachan: @bvellanki, see the pom.xml that the tests are *not* executed in parallel. The change to make the tests dependent on one another is superfluous. 

          <plugin>
                <groupId>org.apache.maven.plugins</groupId>
                <artifactId>maven-surefire-plugin</artifactId>
                <version>2.16</version>
                <configuration>
                    <redirectTestOutputToFile>true</redirectTestOutputToFile>
                    <forkCount>1</forkCount>
                    <reuseForks>false</reuseForks>
                    <forkMode>always</forkMode>
                    <argLine>-Djava.awt.headless=true -Djava.security.krb5.realm= -Djava.security.krb5.kdc=</argLine>
                    <excludedGroups>${excluded.test.groups}</excludedGroups>
                </configuration>
            </plugin>

bvellanki: @vramachan that explains why these tests did not fail in the first place. Thanks for the patch. +1 and will commit soon

yzheng-hortonworks:
+1

PR #187, opened 2016-06-16
FALCON-2027 Enhance documentation on data replication from HDP to Azure

yzheng-hortonworks:
Also fixed a typo in FalconDocumentation.twiki.

### DIFF
--- a/docs/src/site/twiki/DataReplicationAzure.twiki
+++ b/docs/src/site/twiki/DataReplicationAzure.twiki
@@ -1,0 +1,58 @@
+---+ Data Replication between On-premise Hadoop Clusters and Azure Cloud
+
+---++ Overview
+Falcon provides an easy way to replicate data between on-premise Hadoop clusters and Azure cloud.
+With this feature, users would be able to build a hybrid data pipeline,
+e.g. processing sensitive data on-premises for privacy and compliance reasons
+while leverage cloud for elastic scale and online services (e.g. Azure machine learning) with non-sensitive data.
+
+---++ Use Case
+1. Copy data from on-premise Hadoop clusters to Azure cloud
+2. Copy data from Azure cloud to on-premise Hadoop clusters
+
+---++ Usage
+---+++ Set Up Azure Blob Credentials
+To move data to/from Azure blobs, we need to add Azure blob credentials in HDFS.
+This can be done by adding the credential property to core-site.xml directly or through Ambari HDFS configs.
+<verbatim>
+<property>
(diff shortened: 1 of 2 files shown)


docs/src/site/twiki/DataReplicationAzure.twiki
```diff
@@ -0,0 +1,58 @@
...
+
+---++ Overview
+Falcon provides an easy way to replicate data between on-premise Hadoop clusters and Azure cloud.
+With this feature, users would be able to build a hybrid data pipeline,
+e.g. processing sensitive data on-premises for privacy and compliance reasons
+while leverage cloud for elastic scale and online services (e.g. Azure machine learning) with non-sensitive data.
+
+---++ Use Case
+1. Copy data from on-premise Hadoop clusters to Azure cloud
+2. Copy data from Azure cloud to on-premise Hadoop clusters
```
bvellanki: Can you please specify we do not support Azure to Azure ?

docs/src/site/twiki/DataReplicationAzure.twiki
```diff
@@ -0,0 +1,58 @@
...
+
+---++ Usage
+---+++ Set Up Azure Blob Credentials
+To move data to/from Azure blobs, we need to add Azure blob credentials in HDFS.
+This can be done by adding the credential property to core-site.xml directly or through Ambari HDFS configs.
```
bvellanki: If core-site.xml is updated directly, when HDFS is restarted using Ambari - the changes will be lost. So we recommend doing all actions through Ambari if available. Else do all actions using CLI

docs/src/site/twiki/DataReplicationAzure.twiki
yzheng-hortonworks: Good question, @bvellanki . Just tested Azure-to-Azure replication and it works! I added it as one use case.

docs/src/site/twiki/DataReplicationAzure.twiki
yzheng-hortonworks: I reorganized the doc to recommend using Ambari to add Azure blob credentials.

bvellanki:
Very useful doc. +1, will commit shortly

yzheng-hortonworks:
Thank you, @bvellanki .

bvellanki:
The jenkins failure does not seem related to the doc patch submitted. I was able to build without any errors. So I plan to commit the pull request unless there are any objections.

PR #189, opened 2016-06-17
FALCON-2035 Entity list operation without type parameter doesn't work when authorization is enabled

yzheng-hortonworks:
Test entity list operation without type parameter works when authorization is enabled.

### DIFF
--- a/common/src/main/java/org/apache/falcon/security/DefaultAuthorizationProvider.java
+++ b/common/src/main/java/org/apache/falcon/security/DefaultAuthorizationProvider.java
@@ -59,6 +59,7 @@ public class DefaultAuthorizationProvider implements AuthorizationProvider {
 
     private static final Set<String> RESOURCES = new HashSet<String>(
             Arrays.asList(new String[]{"admin", "entities", "instance", "metadata", "extension", }));
+    private static final String LIST_OPERATION = "list";
 
     /**
      * Constant for the configuration property that indicates the prefix.
@@ -170,7 +171,8 @@ public class DefaultAuthorizationProvider implements AuthorizationProvider {
                     authorizeAdminResource(authenticatedUGI, action);
                 }
             } else if ("entities".equals(resource) || "instance".equals(resource)) {
-                authorizeEntityResource(authenticatedUGI, entityName, entityType, action);
+                authorizeEntityResource(authenticatedUGI, entityName, entityType, action,
+                        "entities".equals(resource) && LIST_OPERATION.equals(action));
             } else if ("metadata".equals(resource)) {
                 authorizeMetadataResource(authenticatedUGI, action);
             }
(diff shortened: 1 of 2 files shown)


common/src/main/java/org/apache/falcon/security/DefaultAuthorizationProvider.java
```diff
@@ -170,7 +171,8 @@ public void authorizeResource(String resource, String action,
                     authorizeAdminResource(authenticatedUGI, action);
                 }
             } else if ("entities".equals(resource) || "instance".equals(resource)) {
-                authorizeEntityResource(authenticatedUGI, entityName, entityType, action);
+                authorizeEntityResource(authenticatedUGI, entityName, entityType, action,
+                        "entities".equals(resource) && LIST_OPERATION.equals(action));
```
bvellanki: It is good practice to separate out functional logic from method calls. Have a boolean var that is set to "entities".equals(resource) && LIST_OPERATION.equals(action) and then use it in method call.

common/src/main/java/org/apache/falcon/security/DefaultAuthorizationProvider.java
```diff
@@ -294,15 +296,16 @@ protected boolean isUserInAdminGroups(UserGroupInformation proxyUgi) {
...
-        Validate.notEmpty(entityType, "Entity type cannot be empty or null");
+        if (!entityTypeOptional) {
+            Validate.notEmpty(entityType, "Entity type cannot be empty or null");
+        }
         LOG.debug("Authorizing authenticatedUser={} against entity/instance action={}, "
                 + "entity name={}, entity type={}",
                 authenticatedUGI.getShortUserName(), action, entityName, entityType);
 
-        if (entityName != null) { // lifecycle actions
+        if (entityType != null && entityName != null) { // lifecycle actions
```
bvellanki: This means that there wont be any authorization when entityType is null. Instead, you can change lines 301 to 303 as follows, 

if (entityTypeOptional && entityType == null) {
    Log(validation not required)
    return;
}

What do you think?

common/src/main/java/org/apache/falcon/security/DefaultAuthorizationProvider.java
yzheng-hortonworks: Thank you @bvellanki . I've updated the code.

common/src/main/java/org/apache/falcon/security/DefaultAuthorizationProvider.java
yzheng-hortonworks: @bvellanki We already have logs if this doesn't hold:
{code}
        if (entityType != null && entityName != null) { // lifecycle actions
.......
        } else {
            // non lifecycle actions, lifecycle actions with null entity will validate later
            LOG.info("Authorization for action={} will be done in the API", action);
        }
{code}
We also log the action, entity name, and entity type before this IF check. So by looking at the present log, we would be able to know why we skip authorization for an action that doesn't have type and name information.

bvellanki:
Sounds good to me, +1.

common/src/main/java/org/apache/falcon/security/DefaultAuthorizationProvider.java
```diff
@@ -170,7 +171,8 @@ public void authorizeResource(String resource, String action,
                     authorizeAdminResource(authenticatedUGI, action);
                 }
             } else if ("entities".equals(resource) || "instance".equals(resource)) {
-                authorizeEntityResource(authenticatedUGI, entityName, entityType, action);
+                boolean entityTypeOptional = "entities".equals(resource) && LIST_OPERATION.equals(action);
```
pallavi-rao: Why even modify the authorizeEntityResource method? Authorization does not apply to entity list, with or without type, as entityName won't be available. So, why not:
If (entityTypeOptional) {
     Log("Skipping authorization for list operations")
     return;
}

yzheng-hortonworks: Good point, @pallavi-rao . I moved the "skip" check outside the authorizeEntityResource method.

pallavi-rao:
+1. Will merge if no further comments.

pallavi-rao:
Re-triggered the build and it went through..

PR #190, opened 2016-06-18
FALCON-2036 Update twiki on entity list operation with up-to-date REST API path

yzheng-hortonworks:
1. Mark entity type as optional.
2. Reorganize the order of examples.


pallavi-rao:
+1. Will merge if no further comments. Since a doc change, will merge to both trunk and 0.10

yzheng-hortonworks:
Thank you, @pallavi-rao .

sandeepSamudrala:
:+1:

PR #191, opened 2016-06-20
FALCON-2038 When all Optional input instances are missing, partition should not be suffixed.

pallavi-rao:
Updated code and UT.

### DIFF
--- a/oozie-el-extensions/src/test/java/org/apache/oozie/extensions/TestOozieELExtensions.java
+++ b/oozie-el-extensions/src/test/java/org/apache/oozie/extensions/TestOozieELExtensions.java
@@ -168,6 +168,8 @@ public class TestOozieELExtensions {
                 "*/US", "_DONE", },
             // With availability flag. All instances missing
             {"hdfs://localhost:8020/projects/falcon/staging/EMPTY_DIR_DONT_DELETE", "null", "_FINISH"},
+            // With availability flag and partitions. All instances missing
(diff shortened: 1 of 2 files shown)


oozie-el-extensions/src/test/java/org/apache/oozie/extensions/TestOozieELExtensions.java
```diff
@@ -168,6 +168,8 @@ public void testDataInOptional(String expuris, String partition, String doneFlag
                 "*/US", "_DONE", },
             // With availability flag. All instances missing
             {"hdfs://localhost:8020/projects/falcon/staging/EMPTY_DIR_DONT_DELETE", "null", "_FINISH"},
+            // With availability flag and partitions. All instances missing
+            {"hdfs://localhost:8020/projects/falcon/staging/EMPTY_DIR_DONT_DELETE", "*", "_FINISH"},
```
sandeepSamudrala: what in case of multi level partitions?

pallavi-rao: Code path is the same, whether there is one partition or multi-level. Whether I use * or */*/* in UT, the result will be same.

PraveenAdlakha:
+1

sandeepSamudrala:
:+1:

peeyushb:
Looks good. +1 .

PR #192, opened 2016-06-20
FALCON-2034. Make numThreads and timeOut configurable In ConfigurationStore init

sandeepSamudrala:


src/conf/startup.properties
```diff
@@ -136,6 +136,12 @@ prism.configstore.listeners=org.apache.falcon.entity.v0.EntityGraph,\
 
 *.falcon.cleanup.service.frequency=days(1)
 
+# Default number of threads to be used to restore entities.
```
pallavi-rao: Can we follow the convention we have followed for other variables of config store, such as *.config.store.uri. 
These two could be : 
*.config.store.num.threads.load.entities
*.config.store.start.timeout.minutes

pallavi-rao:
Looks good. +1

src/conf/startup.properties
sandeepSamudrala: Addressed the review comments

src/conf/startup.properties
```diff
@@ -136,6 +136,12 @@ prism.configstore.listeners=org.apache.falcon.entity.v0.EntityGraph,\
...
+*.falcon.default.threads.load.restore.entities=100
+
+# Default timeout in minutes to load entities
+*.falcon.default.timeout.minutes=30
```
peeyushb: Is assigning the 30 minutes timeout reasonable. I am of the opinion it must be less.

sandeepSamudrala: In large organizations/cluster with many feed in the range of 3-4K takes time if the persistent store is hdfs for the prism-store. 
Anyways this is only default value for timeout, if the entities are very few, it will just finish off and restore them

src/conf/startup.properties
peeyushb: thanks for clarification.

peeyushb:
+1.

sandeepSamudrala:
Thanks @pallavi-rao and @peeyushb  for review.

PR #193, opened 2016-06-21
FALCON-2033 ProcessRerun tests fail with User [falcon] not authorized for Coord job

bvellanki:


yzheng-hortonworks:
+1. Better to have a UT or IT for rerun. Could be a separate JIRA :)

oozie/src/main/java/org/apache/falcon/workflow/engine/OozieWorkflowEngine.java
```diff
@@ -1015,7 +1018,7 @@ private String getConsoleUrl(String cluster, String actionId) throws FalconExcep
             // Get conf when workflow is launched.
             if (coordinatorAction.getExternalId() != null) {
                 WorkflowJob jobInfo = client.getJobInfo(coordinatorAction.getExternalId());
-
+                workflowUser = jobInfo.getUser();
```
sowmyaramesh: If you look at  AbstractRerunConsumer it's already taken care of

```
// Login the user to access WfEngine as this user
                CurrentUser.authenticate(message.getWorkflowUser());
```

yzheng-hortonworks: Interesting. Does the message contain the correct workflow user?
From the log, the user for rerun used to be "falcon". Balu's change corrects it.

bvellanki: From the logs reported by @muraliramasami , the rerun user is falcon when the coordinator (thus the workflow) were by user "hrt_qa". So CurrentUser.authenticate(message.getWorkflowUser()) is supposed to work and run re-run as message.getWorkflowUser(), but it is not accurate anymore.  Let me debug some more.

sowmyaramesh: @bvellanki : Can you please verify what message.getWorkflowUser() is if falcon entity is scheduled as a different user than falcon?

If it's falcon then fix you provided should be moved here so that it can handle multiple WF engines : say FalconWorkflowEngine, OozieWorkflowEngine and any other engine supported in future.

bvellanki:
This error was not reproducible in multiple falcon clusters. It is only seen in a specific test cluster, hence we suspect this to be environment issue and closing the Jira. Will reopen if it can be reproduced.

PR #196, opened 2016-06-22
FALCON-2044 Persist Process stats in db

PraveenAdlakha:


PraveenAdlakha:
@pallavi-rao  @bvellanki @sowmyaramesh  Can any one of you please review this ? Its pending for quiet some time now.

prism/src/main/java/org/apache/falcon/plugin/FalconDBNotificationPlugin.java
```diff
@@ -0,0 +1,72 @@
...
+import org.apache.falcon.jdbc.MonitoringJdbcStateStore;
+import org.joda.time.DateTime;
+import org.joda.time.Seconds;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * FalconDB NotificationPlugin.
+ */
+public class FalconDBNotificationPlugin implements MonitoringPlugin {
```
ajayyadava: FalconDBNotificationPlugin doesn't convey much about the functionality of this plugin. Do you think it will be better to rename it to something like "ProcessExecutionStatsPlugin"?

src/conf/startup.properties
```diff
@@ -331,7 +331,8 @@ prism.configstore.listeners=org.apache.falcon.entity.v0.EntityGraph,\
 
 # Setting monitoring plugin, if SMTP parameters is defined
 #*.monitoring.plugins=org.apache.falcon.plugin.DefaultMonitoringPlugin,\
-#                     org.apache.falcon.plugin.EmailNotificationPlugin
+#                     org.apache.falcon.plugin.EmailNotificationPlugin,\
+#                     org.apache.falcon.plugin.FalconDBNotificationPlugin
```
ajayyadava: Although we haven't been doing it till now, I think adding comments around each plugin about the functionality/ scenario where user will want to enable them, dependencies etc. will improve the usability.

prism/src/main/java/org/apache/falcon/plugin/FalconDBNotificationPlugin.java
```diff
@@ -0,0 +1,72 @@
...
+    public void monitor(ResourceMessage message) {
+        try {
+            String entityType = StringUtils.isNotBlank(message.getDimensions().get("entityType"))
+                    ? message.getDimensions().get("entityType") :message.getDimensions().get("entity-type");
+            String entityName = StringUtils.isNotBlank(message.getDimensions().get("entityName"))
+                    ? message.getDimensions().get("entityName") :message.getDimensions().get("entity-name");
+            LOG.debug("message:" + message.getAction());
+            if (entityType.equalsIgnoreCase(EntityType.PROCESS.name())
+                    && ConfigurationStore.get().get(EntityType.PROCESS, entityName) != null) {
+                Entity entity = ConfigurationStore.get().get(EntityType.PROCESS, entityName);
```
ajayyadava: nit: The return value will be of type Process, no need to typecast it.

prism/src/main/java/org/apache/falcon/plugin/FalconDBNotificationPlugin.java
```diff
@@ -0,0 +1,72 @@
...
+                Process process = (Process) entity;
+                String pipeline =  StringUtils.isNotBlank(process.getPipelines()) ? process.getPipelines() : "default";
```
ajayyadava: I suspect it may become an issue if someone has a pipeline named "default".

ajayyadava: As of now a process can belong to multiple pipelines (comma separated value). In such cases it will become difficult to query using a single pipeline or get stats for a pipeline. One possible solution can be to store an array of values and query for pipeline name in that array. I know that at least postgres has support for array types.

src/conf/startup.properties
PraveenAdlakha: We can add one line comments in the properties file but the functionality and scenario should go the twiki.Let me know your thoughts will work accordingly.

prism/src/main/java/org/apache/falcon/plugin/FalconDBNotificationPlugin.java
PraveenAdlakha: If pipeline name is default then default will go as pipeline tag. Am I missing anything here?

Regarding comma separated value in pipeline tag I think we should write multiple entries for the process each with different pipeline tag.So that anyone who is querying the db on pipeline level will get all the processes associated with it separately.

Let me know your thoughts on this one too will make changes accordingly..

ajayyadava:
@PraveenAdlakha Can you please rebase and run jenkins again(close and open the pull request)?

ajayyadava:
👍 Will commit shortly

PR #203, opened 2016-06-30
FALCON-2057 HiveDR not working with multiple users and same DB

bvellanki:


yzheng-hortonworks:
+1. Could you add documentation that falcon will create baseDir and user will create dbName and jobName? Could you also document that the user needs to be added to users group in order to run DR jobs? Thank you, @bvellanki .

bvellanki:
@yzheng-hortonworks  This jira is not creating new behavior. It is fixing a bug in existing behavior. The users should not be aware of who creates what dir because the data stored under these dirs is "status of replication". This should be transparent to users.

yzheng-hortonworks:
Could you add a comment to the code? So other developers would be aware of this? Thank you!

vrangan:
+1

bvellanki:
@yzheng-hortonworks  : I added comments  in code. Thank you for the review

sowmyaramesh:
+1

PR #205, opened 2016-07-01
FALCON-1914 Hive mirroring should work for secure to unsecure & viceversa. Also cleanup artifacts.

sowmyaramesh:


bvellanki:
+1 looks good to me

PR #206, opened 2016-07-01
FALCON-2058 s3 tests with dummy url no longer compatible with latest HDFS

yzheng-hortonworks:
With latest HDFS 2.7.2 or above, falcon's validation with dummy s3 url would fail. Therefore disabling this s3 replication UT. Valid s3 case is already covered in regression tests.

### DIFF
--- a/common/src/test/java/org/apache/falcon/entity/parser/FeedEntityParserTest.java
+++ b/common/src/test/java/org/apache/falcon/entity/parser/FeedEntityParserTest.java
@@ -978,6 +978,7 @@ public class FeedEntityParserTest extends AbstractTestBase {
     }
 
     // disable this test due to its validation of dummy s3 url no longer supported by latest hdfs (2.7.2 or above)
+    @Test (enabled = false)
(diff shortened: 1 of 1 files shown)


common/src/test/java/org/apache/falcon/entity/parser/FeedEntityParserTest.java
```diff
@@ -977,7 +977,7 @@ private void deleteLocations(Feed feed) throws IOException {
         }
     }
 
-    @Test
+    // disable this test due to its validation of dummy s3 url no longer supported by latest hdfs (2.7.2 or above)
```
peeyushb: To disable test , can we use @Test (enabled = false) along with comment.

yzheng-hortonworks: Good point. Updated. Thank you for your review, @peeyushb .

bvellanki:
+1

PR #207, opened 2016-07-01
FALCON-2056 HiveDR doesn't work with multiple users

sowmyaramesh:


sowmyaramesh:
Build failure is not related to the patch, rebuilding.

bvellanki:
+1 @sowmyaramesh , will commit today if there are no further comments.

peeyushb:
+1. Looks good to me.

addons/hivedr/src/main/java/org/apache/falcon/hive/util/HiveDRStatusStore.java
```diff
@@ -49,7 +49,10 @@
     private static final Logger LOG = LoggerFactory.getLogger(DRStatusStore.class);
     private FileSystem fileSystem;
 
-    private static final String DEFAULT_STORE_PATH = BASE_DEFAULT_STORE_PATH + "hiveReplicationStatusStore/";
+    private static final String DEFAULT_STORE_PATH = StringUtils.removeEnd
+            (DRStatusStore.BASE_DEFAULT_STORE_PATH,  File.separator) + File.separator
```
yzheng-hortonworks: I see you try to append File.separator if it is not at the end of the string. StringUtils::appendIfMissing looks better in this case.

addons/hivedr/src/main/java/org/apache/falcon/hive/util/FileUtils.java
```diff
@@ -34,8 +34,10 @@
 public final class FileUtils {
 
     public static final String DEFAULT_EVENT_STORE_PATH = StringUtils.removeEnd(DRStatusStore
-            .BASE_DEFAULT_STORE_PATH,  File.separator) + File.separator + "Events";
+            .BASE_DEFAULT_STORE_PATH,  File.separator) + File.separator + "Events" + File.separator;
```
yzheng-hortonworks: I see you try to append File.separator if it is not at the end of the string. StringUtils::appendIfMissing would be better to use in this case. Same for the other place in

bvellanki: @yzheng-hortonworks  That is a good point. Since it is a single line change, I will change it while commit to unblock 0.10

PR #211, opened 2016-07-01
FALCON-2050: Configure jetty parent classloader to be prioritized ove…

vrangan:
Enabled the parent classloader to have priority over the webapp classloader


bvellanki:
+1 One line change, so will commit now

PR #212, opened 2016-07-06
FALCON-2059 BacklogMetricEmitter Service for Falcon Processes

pavankumar526:


common/src/main/java/org/apache/falcon/persistence/BacklogMetricBean.java
```diff
@@ -0,0 +1,116 @@
...
+        @NamedQuery(name = PersistenceConstants.DELETE_BACKLOG_METRIC_INSTANCE, query = "delete from BacklogMetricBean a where a.entityName = :entityName and a.clusterName = :clusterName and a.nominalTime = :nominalTime and a.entityType = :entityType")
+})
+//RESUME CHECKSTYLE CHECK  LineLengthCheck
+
+@Table(name = "BACKLOG_METRIC")
+public class BacklogMetricBean {
+
+    @NotNull
+    @GeneratedValue(strategy = GenerationType.AUTO)
+    @Id
```
pallavi-rao: Why is this required?

common/src/test/java/org/apache/falcon/entity/parser/ProcessEntityParserTest.java
```diff
@@ -103,7 +103,7 @@ public void testParse() throws FalconException, JAXBException {
 
         Assert.assertEquals(process.getTags(),
                 "consumer=consumer@xyz.com, owner=producer@xyz.com, _department_type=forecasting");
-        Assert.assertEquals(process.getPipelines(), "testPipeline,dataReplication_Pipeline");
+        Assert.assertEquals(process.getPipelines(), "testPipeline");
```
pallavi-rao: Why this change?

prism/src/main/java/org/apache/falcon/jdbc/BacklogMetricStore.java
```diff
@@ -0,0 +1,121 @@
...
+        EntityManager entityManager = getEntityManager();
+        try {
+            beginTransaction(entityManager);
+            entityManager.persist(backlogMetricBean);
+        } finally {
+            commitAndCloseTransaction(entityManager);
+        }
+    }
+
+    public synchronized void deleteMonitoringFeed(String entityName, String cluster, Date nominalTime,
```
pallavi-rao: Change method name

prism/src/main/java/org/apache/falcon/service/BacklogMetricEmitterService.java
```diff
@@ -0,0 +1,313 @@
...
+    }
+
+    private BacklogMetricEmitterService() {
+    }
+
+    private ScheduledThreadPoolExecutor scheduledThreadPoolExecutor1 = new ScheduledThreadPoolExecutor(1);
+    private ScheduledThreadPoolExecutor scheduledThreadPoolExecutor2 = new ScheduledThreadPoolExecutor(1);
+
+
+    public static final ThreadLocal<SimpleDateFormat> DATE_FORMAT = new ThreadLocal<SimpleDateFormat>() {
```
pallavi-rao: Why is this ThreadLocal?

prism/src/main/java/org/apache/falcon/service/BacklogMetricEmitterService.java
```diff
@@ -0,0 +1,313 @@
...
+
+/**
+ * Backlog Metric Emitter Service to publish metrics to Graphite.
+ */
+public final class BacklogMetricEmitterService implements FalconService,
+        EntitySLAListener, WorkflowExecutionListener {
+
+    private static final String METRIC_PREFIX = "falcon";
+    private static final String METRIC_SEPARATOR = ".";
+    private static final String BACKLOG_METRIC_EMIT_INTERVAL = "falcon.backlog.metricservice.emit.interval.millisecs";
```
pallavi-rao: millisecs is too granular. Use seconds instead.

prism/src/main/java/org/apache/falcon/service/BacklogMetricEmitterService.java
```diff
@@ -0,0 +1,313 @@
...
+                "60000"));
+        int recheckInterval = Integer.parseInt(StartupProperties.get().getProperty(BACKLOG_METRIC_RECHECK_INTERVAL,
+                "60000"));
+        scheduledThreadPoolExecutor1.scheduleAtFixedRate(new BacklogMetricEmitter(),
+                1, emitInterval, TimeUnit.MILLISECONDS);
+        scheduledThreadPoolExecutor2.scheduleAtFixedRate(new BacklogCheckService(),
+                1, recheckInterval, TimeUnit.MILLISECONDS);
+    }
+
+    private void initInstances() throws FalconException {
```
pallavi-rao: If I'm not wrong, this service will run on Falcon Server and not Prism. In that case, shouldn't this server only initialize o nly with Entities that are scheduled on that cluster, and not all entities?

prism/src/main/java/org/apache/falcon/service/BacklogMetricEmitterService.java
```diff
@@ -0,0 +1,313 @@
...
+            long backlog = 0;
+            MetricInfo metricInfo = null;
+            synchronized (metrics) {
+                long currentTime = System.currentTimeMillis();
+                Iterator iter = metrics.iterator();
+                while (iter.hasNext()) {
+                    try {
+                        metricInfo = (MetricInfo) iter.next();
+                        long time = DATE_FORMAT.get().parse(metricInfo.getNominalTime()).getTime();
+                        backlog += (currentTime - time);
```
pallavi-rao: Backlog calculation has to be per cluster. In this loop, you are not differentiating between clusters. Theoretically, one Falcon Server can manage multiple clusters.

prism/src/main/java/org/apache/falcon/service/BacklogMetricEmitterService.java
```diff
@@ -0,0 +1,314 @@
...
+            backlog = backlog / (60 * 1000); // Converting to minutes
+            metricNotificationService.publish(metricName, backlog);
+        }
+    }
+
+
+    /**
+     * Service runs periodically and removes succeeded instances from backlog list.
+     */
+    public static class BacklogCheckService implements Runnable {
```
pallavi-rao: Can we merge this service with the BacklogCalcService. Add to the backlog only when the instance has still not succeeded?

pallavi-rao:
Add to documentation along with this pull request or file a separate JIRA and handle it in a separate PR.

common/src/main/java/org/apache/falcon/persistence/BacklogMetricBean.java
pavankumar526: Not required for usage, added since there is no primary key in the table.

common/src/test/java/org/apache/falcon/entity/parser/ProcessEntityParserTest.java
pavankumar526: I have an assumption that one process can belong to only one pipeline but feed can belong to multiple pipelines.

prism/src/main/java/org/apache/falcon/jdbc/BacklogMetricStore.java
pavankumar526: sure will change

prism/src/main/java/org/apache/falcon/service/BacklogMetricEmitterService.java
pavankumar526: Since for unit tests i made it like this. Otherwise tests will take longer .

prism/src/main/java/org/apache/falcon/service/BacklogMetricEmitterService.java
pavankumar526: Sure will change

prism/src/main/java/org/apache/falcon/service/BacklogMetricEmitterService.java
pavankumar526: This will be useful in few cases like when we get success notification and deleted entry from state store but before deleting from memory if falcon was restarted. Then that stale will instance will stuck forever this service will be useful to remove it from memory.

prism/src/test/resources/startup.properties
```diff
@@ -0,0 +1,338 @@
+#
```
pavankumar526: sure

prism/src/main/java/org/apache/falcon/service/BacklogMetricEmitterService.java
pavankumar526: I am not using more features except parsing and formatting,  and also simple date format used lot of places in falcon.

common/src/test/java/org/apache/falcon/entity/parser/ProcessEntityParserTest.java
pallavi-rao: pipeline does not apply to a feed at all. Per process definition, a process can have multiple definitions. So, you should not make that assumption. Emit a metric per pipeline, if that is what is required.

pallavi-rao:
All changes look good, except that assumption that the pipeline element will have only 1 pipeline name.

pavankumar526:
@pallavi-rao  I have addressed multiple pipelines issue. Can you please check now.

pallavi-rao:
+1. Will merge on the morrow.

common/src/main/java/org/apache/falcon/persistence/BacklogMetricBean.java
PraveenAdlakha: its good to have integer as primary key and we are following it in the same way every where in falcon.

common/src/main/resources/META-INF/persistence.xml
```diff
@@ -91,7 +92,7 @@
         <class>org.apache.falcon.persistence.MonitoredEntityBean</class>
         <class>org.apache.falcon.persistence.PendingInstanceBean</class>
         <class>org.apache.falcon.persistence.EntitySLAAlertBean</class>
-
+        <class>org.apache.falcon.persistence.BacklogMetricBean</class>
```
PraveenAdlakha: please add the class name every where currently its missing in mysql and derby in tag openjpa.MetaDataFactory.

prism/src/main/java/org/apache/falcon/service/BacklogMetricEmitterService.java
PraveenAdlakha: Lets stick with seconds and we can play around with the default value for the tests.

pallavi-rao: +1 to @PraveenAdlakha's suggestion. We can set the default values while running UTs.

prism/src/main/java/org/apache/falcon/service/BacklogMetricEmitterService.java
```diff
@@ -0,0 +1,355 @@
...
+                                String nominalTimeStr = metricInfo.getNominalTime();
+                                Date nominalTime;
+                                try {
+                                    nominalTime = DATE_FORMAT.get().parse(nominalTimeStr);
+                                    if (entity.getACL().getOwner() != null && !entity.getACL().getOwner().isEmpty()) {
+                                        CurrentUser.authenticate(entity.getACL().getOwner());
+                                    } else {
+                                        CurrentUser.authenticate(System.getProperty("user.name"));
+                                    }
+                                    if (wfEngine.isMissing(entity)) {
```
pallavi-rao: Better listen to ConfigStore changes (onDelete) instead.

prism/src/main/java/org/apache/falcon/service/BacklogMetricEmitterService.java
```diff
@@ -0,0 +1,355 @@
...
+    /**
+     * Service which executes backlog evaluation and publishing metrics to Graphite parallel for entities.
+     */
+    public static class BacklogMetricEmitter implements Runnable {
+        private ThreadPoolExecutor executor;
+
+        @Override
+        public void run() {
+            LOG.debug("BacklogMetricEmitter running for entities");
+            executor = new ScheduledThreadPoolExecutor(10);
```
PraveenAdlakha: Lets keep the thread number higher around 100 as the for loop will make only use as per requirement.

prism/src/main/java/org/apache/falcon/service/BacklogMetricEmitterService.java
pavankumar526: I tried doing that but i found few problems in that approach, Once we have found entity was deleted , we have to store that entity in a list and check whether instances are coming or not. Until what time we have to maintain the list ? 
What if falcon restart happened after that entity was deleted and list will be cleared and we got instances from that entity saying instance missed sla high ?

pallavi-rao:
@pavankumar526, can you please fix the checkstyle issues (that is causing the build failure)?

pallavi-rao:
The build has gone through fine. Merging.

PR #213, opened 2016-07-07
FALCON-2061 Falcon CLI shows hadoop classpath loading info in the console

bvellanki:


vrangan:
+1

src/bin/falcon-config.sh
```diff
@@ -68,18 +68,6 @@ case $type in
     for i in `ls ${BASEDIR}/server/webapp`; do
       FALCONCPPATH="${FALCONCPPATH}:${i}/WEB-INF/lib/*"
     done
-    # In future we should refactor the CP directories between server and client
-    HADOOPDIR=`which hadoop`
-    if [ "$HADOOP_HOME" != "" ]; then
-      echo "Hadoop home is set, adding libraries from '${HADOOP_HOME}/bin/hadoop classpath' into falcon classpath"
-      FALCONCPPATH="${FALCONCPPATH}:`${HADOOP_HOME}/bin/hadoop classpath`"
```
yzheng-hortonworks: Should the update of FALCONCPPATH be kept here? It is used in later part of the code. Same for the one in "elif".

vramachan:
@bvellanki falcon-config.sh is used by the service-start in addition to falcon cli client. won't the Hadoop libs needed for the server start?

vramachan:
I see you are making the change in the client APPTYPE.
+1

src/bin/falcon-config.sh
bvellanki: The change is only made in Client side. All the jars required by client are in client lib and in server/webapp/web-inf/lib/

yzheng-hortonworks:
+1

PR #214, opened 2016-07-07
FALCON-2030 : Enfornce time partition in the storage path for feeds w…

vramachan:
Enforce time partition in the storage path for feeds with import and export policy.
HCAT storage types, this restriction is already handled.

### DIFF
--- a/common/src/test/java/org/apache/falcon/entity/parser/FeedEntityParserTest.java
+++ b/common/src/test/java/org/apache/falcon/entity/parser/FeedEntityParserTest.java
@@ -1160,7 +1160,7 @@ public class FeedEntityParserTest extends AbstractTestBase {
         Assert.fail("ValidationException should have been thrown");
     }
 
-    @Test (expectedExceptions = {ValidationException.class, FalconException.class})
+    @Test (expectedExceptions = {ValidationException.class})
     public void testImportFeedWithNoTimePartition() throws Exception {
 
         InputStream feedStream = this.getClass()
@@ -1169,7 +1169,7 @@ public class FeedEntityParserTest extends AbstractTestBase {
         Assert.fail("ValidationException should have been thrown");
     }
 
-    @Test (expectedExceptions = {ValidationException.class, FalconException.class})
+    @Test (expectedExceptions = {ValidationException.class})
     public void testImportFeedWithInvalidTimePartition() throws Exception {
 
         InputStream feedStream = this.getClass()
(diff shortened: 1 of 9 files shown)


common/src/test/java/org/apache/falcon/entity/parser/FeedEntityParserTest.java
```diff
@@ -1160,6 +1160,24 @@ public void testImportFeedSqoopInvalid() throws Exception {
         Assert.fail("ValidationException should have been thrown");
     }
 
+    @Test (expectedExceptions = {ValidationException.class, FalconException.class})
+    public void testImportFeedWithNoTimePartition() throws Exception {
+
+        InputStream feedStream = this.getClass()
+                .getResourceAsStream("/config/feed/feed-import-no-timepartition-0.1.xml");
+        parser.parseAndValidate(feedStream);
+        Assert.fail("ValidationException should have been thrown");
```
ajayyadava: This should already be handled by expectedExceptions, right?

common/src/test/java/org/apache/falcon/entity/parser/FeedEntityParserTest.java
```diff
@@ -1160,6 +1160,24 @@ public void testImportFeedSqoopInvalid() throws Exception {
...
+    }
+
+    @Test (expectedExceptions = {ValidationException.class, FalconException.class})
```
ajayyadava: Shouldn't this be just ValidationException.class. Using two exceptions make the test a bit indeterministic and may pass in other cases as well.

ajayyadava:
It will be useful to mention it in the docs( entity specification of feed) that the pattern is compulsory for the following scenarios along with some examples of valid and invalid patterns.

common/src/test/java/org/apache/falcon/entity/parser/FeedEntityParserTest.java
vramachan: @ajayyadava fixed expected exceptions and added documentation.

yzheng-hortonworks:
+1

common/src/test/java/org/apache/falcon/entity/parser/FeedEntityParserTest.java
ajayyadava: Hey @vramachan Thanks for fixing the docs and expected exceptions. Actually in this statement I meant something else. I am curious about the need for that Assert.fail statement. Even in failure scenarios, even without this statement we will get clear message that the ValidationException was expected but was not thrown. Anyhow, it is just nit picking :stuck_out_tongue:

ajayyadava:
LGTM :+1:

PR #218, opened 2016-07-08
FALCON-2065 : Falcon CLI failing with missing KerberosAuthenticator class

peeyushb:


src/bin/falcon-config.sh
```diff
@@ -68,6 +68,18 @@ case $type in
     for i in `ls ${BASEDIR}/server/webapp`; do
       FALCONCPPATH="${FALCONCPPATH}:${i}/WEB-INF/lib/*"
     done
+    # In future we should refactor the CP directories between server and client
```
bvellanki: Instead of adding back the `hadoop classpath` into client classpath, can you please provide a new fix that will do the following,
- Remove the code in falcon-config.sh that adds `hadoop classpath` to falcon client class path 
- make sure <falcon home>/client/lib/ has following jar files
hadoop-auth-<version>.jar    hadoop-hdfs-<version>.jar  hadoop-client-<version>.jar
This can be achieved by removing scope “provided" for these jars in pom.xml

bvellanki:
👍  thanks @peeyushb

PR #222, opened 2016-07-12
FALCON-1749 Instance status does not show instances if entity is dele…

PraveenAdlakha:
…ted from one of the colos

### DIFF
--- a/prism/src/main/java/org/apache/falcon/resource/AbstractEntityManager.java
+++ b/prism/src/main/java/org/apache/falcon/resource/AbstractEntityManager.java
@@ -25,7 +25,6 @@ import org.apache.falcon.FalconException;
 import org.apache.falcon.FalconRuntimException;
 import org.apache.falcon.FalconWebException;
 import org.apache.falcon.Pair;
-import org.apache.falcon.entity.ClusterHelper;
 import org.apache.falcon.entity.EntityNotRegisteredException;
 import org.apache.falcon.entity.EntityUtil;
 import org.apache.falcon.entity.lock.MemoryLocks;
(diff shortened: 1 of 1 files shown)


prism/src/main/java/org/apache/falcon/resource/AbstractEntityManager.java
```diff
@@ -172,6 +173,9 @@ protected static void checkColo(String colo) {
             for (String cluster : clusters) {
                 try{
                     Cluster clusterEntity = EntityUtil.getEntity(EntityType.CLUSTER, cluster);
+                    if(type.equals(EntityType.PROCESS)){
+                        EntityUtil.getEntity(EntityType.PROCESS,entity.getName());
```
bvellanki: What does this code achieve? I do not understand how this will fix the issue, please elaborate.

PraveenAdlakha: Hi,

The root cause of the issue was the process was present on the prism but was deleted from one of the colo's in distributed mode and we are getting a NPE for that colo when we do a status. 
Through this code I am checking in the configrationstore if that process actually exists at that colo.As for one of the clusters it will give me EntityNotRegisteredException, we are catching it and we will not get NPE.

bvellanki: Ah, that makes sense. Please add a comment here for future reference.

bvellanki:
+1

prism/src/main/java/org/apache/falcon/resource/AbstractEntityManager.java
```diff
@@ -25,6 +25,7 @@
 import org.apache.falcon.FalconRuntimException;
 import org.apache.falcon.FalconWebException;
 import org.apache.falcon.Pair;
+import org.apache.falcon.entity.ClusterHelper;
```
pallavi-rao: This import is not used in your code change

prism/src/main/java/org/apache/falcon/resource/AbstractEntityManager.java
```diff
@@ -172,6 +173,10 @@ protected static void checkColo(String colo) {
             for (String cluster : clusters) {
                 try{
                     Cluster clusterEntity = EntityUtil.getEntity(EntityType.CLUSTER, cluster);
+                    // Fix for Falcon-1749
+                    if(type.equals(EntityType.PROCESS)){
```
pallavi-rao: What if it is a feed? Why only process?

pallavi-rao:
+1. The build failure seems unable. Will merge.

PR #223, opened 2016-07-12
FALCON-2073 Handle with NULL corner case

yzheng-hortonworks:
Changes:
1. EntityGraph::getDependents should return empty list instead of NULL value if there is no dependent entities. Affected method include OozieWorkflowEngine::updateDependant and AbstractEntityManager::getDependencies. Also removed unnecessary existing NULL checks after the change.
2. In LogProvider::populateActionLogUrls, handle with the NULL case where there is no file status under the specified path.
3. FalconClient::getDependency should return empty list instead of NULL value if there is no dependent entities. Affected method include FalconEntityCLI::entityCommand.

Extra minor changes in this patch:
4. A regular expression misusage in LogProvider::getActionName. To match special character ".", should use "[.]" instead of "." which will try to match any character.
5. Fix a rat check error in Falcon CLI due to the introduction of Spring shell commands.

### DIFF
--- a/client/src/main/java/org/apache/falcon/client/FalconClient.java
+++ b/client/src/main/java/org/apache/falcon/client/FalconClient.java
@@ -470,7 +470,7 @@ public class FalconClient extends AbstractFalconClient {
         checkIfSuccessful(clientResponse);
 
         EntityList result = clientResponse.getEntity(EntityList.class);
-        if (result == null || result.getElements() == null) {
+        if (result == null) {
             return new EntityList();
         }
         return result;
--- a/client/src/main/java/org/apache/falcon/client/FalconClient.java
+++ b/client/src/main/java/org/apache/falcon/client/FalconClient.java
@@ -471,7 +471,7 @@ public class FalconClient extends AbstractFalconClient {
 
         EntityList result = clientResponse.getEntity(EntityList.class);
         if (result == null || result.getElements() == null) {
-            return null;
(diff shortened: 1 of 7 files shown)


client/src/main/java/org/apache/falcon/client/FalconClient.java
```diff
@@ -471,7 +471,7 @@ public EntityList getDependency(String entityType, String entityName, String doA
 
         EntityList result = clientResponse.getEntity(EntityList.class);
         if (result == null || result.getElements() == null) {
```
bvellanki: You dont need the " result.getElements() == null " condition

bvellanki:
Apart from minor nit, +1

client/src/main/java/org/apache/falcon/client/FalconClient.java
yzheng-hortonworks: Removed. Thanks @bvellanki for review.

peeyushb:
Patch looks good. +1 .

vrangan:
+1  Thanks @yzheng-hortonworks   please mention that these were from the coverity scan to give due credit for finding the issues

yzheng-hortonworks:
@vrangan Thanks for review. It is mentioned in the JIRA already :)

peeyushb:
If no further review comments, I will commit this patch soon.

PR #227, opened 2016-07-14
FALCON-2080 Exceptions in the log

PraveenAdlakha:


prism/src/main/java/org/apache/falcon/service/EntitySLAMonitoringService.java
```diff
@@ -442,7 +443,11 @@ private boolean checkEntityInstanceAvailability(String entityName, String cluste
                                                     String entityType) throws
         FalconException {
         Entity entity = EntityUtil.getEntity(entityType, entityName);
-
+        if (entity.getACL().getOwner() != null && !entity.getACL().getOwner().isEmpty()) {
```
bvellanki: It might be better to have a method like OozieWorkflowEngine::switchUser() and switch back after the checkInstance command is run.

peeyushb: StringUtils.isNotBlank() can be used, instead of checking for null and empty condition for owner String value.

pallavi-rao:
Agree with @bvellanki.  I too think it is better to use the switchUser in OozieWorkflowEngine like it is done in other methods of OWE.

prism/src/main/java/org/apache/falcon/service/EntitySLAMonitoringService.java
```diff
@@ -439,10 +441,9 @@ private void checkPendingInstanceAvailability(String entityType) throws FalconEx
 
     // checks whether a given feed instance is available or not
     private boolean checkEntityInstanceAvailability(String entityName, String clusterName, Date nominalTime,
-                                                    String entityType) throws
-        FalconException {
+                                                    String entityType) throws FalconException {
         Entity entity = EntityUtil.getEntity(entityType, entityName);
-
+        autheticateUser(entity);
```
pallavi-rao: Nit: Typo : authenticate (it says autheticate)

PraveenAdlakha:
Hi All,

As we will be running the code in multithreaded mode thats why not removed the static method which was there in earlier commit.
Also I am not using OWE method as both the methods are not exactly same.Let me know in case there is any concern.

Thanks

pallavi-rao:
+1

peeyushb:
+1. Looks good.

PR #228, opened 2016-07-15
FALCON-2060 Retry does not happen if instance timedout

pallavi-rao:
These got broken when we moved from wf rerun to coord rerun.

### DIFF
--- a/oozie/src/main/java/org/apache/falcon/workflow/engine/OozieWorkflowEngine.java
+++ b/oozie/src/main/java/org/apache/falcon/workflow/engine/OozieWorkflowEngine.java
@@ -1603,6 +1603,8 @@ public class OozieWorkflowEngine extends AbstractWorkflowEngine {
             } else if (jobId.endsWith("-B")) {
                 BundleJob bundle = client.getBundleJobInfo(jobId);
                 return bundle.getStatus().name();
+            } else if (jobId.contains("-C@")) {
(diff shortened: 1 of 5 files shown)


oozie/src/main/java/org/apache/falcon/workflow/engine/OozieWorkflowEngine.java
```diff
@@ -1603,6 +1603,8 @@ public String getWorkflowStatus(String cluster, String jobId) throws FalconExcep
             } else if (jobId.endsWith("-B")) {
                 BundleJob bundle = client.getBundleJobInfo(jobId);
                 return bundle.getStatus().name();
+            } else if (jobId.contains("-C@")) {
```
peeyushb: Can jobId contains "-C@", I think coordinator action/instance id contains "-C@" . Please confirm.

pallavi-rao: jobId is just a variable name. We are passing bundles, coordinators, workflows and coord actions to this method. Yes. -C@ is applicable only for coord action instance.

peeyushb:
Thanks @pallavi-rao for clarification. 

Patch looks good. +1.

sandeepSamudrala:
:+1:

pallavi-rao:
Thanks @peeyushb, @sandeepSamudrala for the review. Merging as this is a blocker for 0.10.

PR #231, opened 2016-07-19
Falcon 2084:  HCatReplicationTest are failing in secure mode

vrangan:
Removed the dependency on job-xml whose priority has changed because of some backward incompatible changes in Oozie - this patch will work with 4.2.0 and upcoming 4.3.0


bvellanki:
@vrangan  : Thank you for fixing and testing this. This bug seems important enough to go into 0.10, please update fix-version for Jira if you agree.

+1 for the patch.

vrangan:
Yes.   Please commit to 0.10

oozie/src/main/java/org/apache/falcon/oozie/feed/HCatReplicationWorkflowBuilder.java
```diff
@@ -72,6 +84,16 @@ public HCatReplicationWorkflowBuilder(Feed entity) {
             OozieUtils.unMarshalHiveAction(export);
         org.apache.falcon.oozie.hive.ACTION hiveExportAction = exportActionJaxbElement.getValue();
         addHDFSServersConfig(hiveExportAction, src, target);
+        Properties hiveConf = HiveUtil.getHiveCredentials(src);
```
yzheng-hortonworks: This looks similar to the code you added at line 66-74. Better to organize them as one function that propagates hive credential to an action.

oozie/src/main/java/org/apache/falcon/oozie/feed/HCatReplicationWorkflowBuilder.java
```diff
@@ -133,8 +165,9 @@ private void setupHiveCredentials(Cluster sourceCluster, Cluster targetCluster,
                 (org.apache.falcon.oozie.workflow.ACTION) object;
             String actionName = action.getName();
             if (PREPROCESS_ACTION_NAME.equals(actionName)) {
+
                 // add reference to hive-site conf to each action
-                action.getJava().setJobXml("${wf:appPath()}/conf/falcon-source-hive-site.xml");
+                //action.getJava().setJobXml("${wf:appPath()}/conf/falcon-source-hive-site.xml");
```
yzheng-hortonworks: Any reason to comment out this line instead of deleting line 169-170?

oozie/src/main/java/org/apache/falcon/oozie/feed/HCatReplicationWorkflowBuilder.java
vrangan: The action types are different for preprocess and hive actions with different classes - You can't pass a single base type to them.   But for Hive actions we can do a single function, but even there since we print action type specific logs, it will be a little bit of an overkill.   We can clean it up later

oozie/src/main/java/org/apache/falcon/oozie/feed/HCatReplicationWorkflowBuilder.java
vrangan: We can remove this.  Will do

yzheng-hortonworks:
+1

PR #234, opened 2016-07-20
Falcon 2083 - Support for arguments (like direct mode and verbose) in database import and export

vramachan:


common/src/main/java/org/apache/falcon/entity/parser/FeedEntityParser.java
```diff
@@ -33,18 +33,7 @@
...
-import org.apache.falcon.entity.v0.feed.Feed;
-import org.apache.falcon.entity.v0.feed.Cluster;
-import org.apache.falcon.entity.v0.feed.ClusterType;
-import org.apache.falcon.entity.v0.feed.Location;
-import org.apache.falcon.entity.v0.feed.LocationType;
-import org.apache.falcon.entity.v0.feed.MergeType;
-import org.apache.falcon.entity.v0.feed.Properties;
-import org.apache.falcon.entity.v0.feed.Property;
-import org.apache.falcon.entity.v0.feed.Sla;
+import org.apache.falcon.entity.v0.feed.*;
```
PraveenAdlakha: this will cause a checkstyle failure.

oozie/src/main/java/org/apache/falcon/oozie/DatabaseImportWorkflowBuilder.java
```diff
@@ -134,15 +134,57 @@ private StringBuilder buildTargetDirArg(StringBuilder builder)
...
-            builder.append(e.getKey()).append(ImportExportCommon.ARG_SEPARATOR).append(e.getValue())
-                .append(ImportExportCommon.ARG_SEPARATOR);
-        }
+        ImportExportCommon.handleVerbose(builder, extraArgs);
+        ImportExportCommon.buildArguments(builder, extraArgs);
+        ImportExportCommon.handleDirectMode(builder, extraArgs);
         return builder;
     }
 
+//    private StringBuilder handleVerbose(StringBuilder builder, Map<String, String> extraArgs)
```
PraveenAdlakha: please remove the commented code.

oozie/src/main/java/org/apache/falcon/oozie/ImportExportCommon.java
```diff
@@ -129,4 +129,45 @@ public static StringBuilder buildTableArg(StringBuilder builder, String tableNam
...
+        for(Map.Entry<String, String> e : extraArgs.entrySet()) {
+            if (!e.getKey().startsWith("--direct")) {
+                builder.append(e.getKey()).append(ImportExportCommon.ARG_SEPARATOR).append(e.getValue())
+                        .append(ImportExportCommon.ARG_SEPARATOR);
+            }
+        }
+        return builder;
+    }
+
+    public static StringBuilder handleVerbose(StringBuilder builder, Map<String, String> extraArgs)
```
PraveenAdlakha: please change the method name to something like set direct rather than handle...handle looks misleading.

oozie/src/main/java/org/apache/falcon/oozie/ImportExportCommon.java
```diff
@@ -129,4 +129,45 @@ public static StringBuilder buildTableArg(StringBuilder builder, String tableNam
...
+            throws FalconException {
+        if ((extraArgs.containsKey("--verbose")) && (extraArgs.get("--verbose").equalsIgnoreCase("TRUE"))) {
+            builder.append("--verbose").append(ImportExportCommon.ARG_SEPARATOR);
+            extraArgs.remove("--verbose");
+        }
+        return builder;
```
PraveenAdlakha: return type is not required here as we are using the same builder received in the input.

oozie/src/main/java/org/apache/falcon/oozie/ImportExportCommon.java
```diff
@@ -129,4 +129,45 @@ public static StringBuilder buildTableArg(StringBuilder builder, String tableNam
...
+    public static StringBuilder handleDirectMode(StringBuilder builder, Map<String, String> extraArgs)
+            throws FalconException {
+        if ((extraArgs.containsKey("--direct")) && (extraArgs.get("--direct").equalsIgnoreCase("true"))) {
+            builder.append("--direct").append(ImportExportCommon.ARG_SEPARATOR);
+        }
+        boolean first = true;
+        for(Map.Entry<String, String> e : extraArgs.entrySet()) {
+            if (e.getKey().startsWith("--direct.")) {
+                if (first) {
+                    builder.append("--").append(ImportExportCommon.ARG_SEPARATOR);
```
PraveenAdlakha: we have already done this check above...Is this intentional?

Please change the method name here also.Can you please think of another way or separate out the case in which we are handling other arguments in handleDirectMode as per the name it should not be playing around with rest of the arguments. Code will be more cleaner and readable that way.

oozie/src/main/java/org/apache/falcon/oozie/ImportExportCommon.java
vramachan: This will be needed if the methods are chained together.

oozie/src/main/java/org/apache/falcon/oozie/ImportExportCommon.java
vramachan: it's handling all the arguments related to direct mode.

oozie/src/main/java/org/apache/falcon/oozie/ImportExportCommon.java
vramachan: I don't think so, but it can be handleVerboseArg.

oozie/src/main/java/org/apache/falcon/oozie/ImportExportCommon.java
vrangan: Why do we add -- before the direct arg?

common/src/main/java/org/apache/falcon/entity/parser/FeedEntityParser.java
vramachan: checkstyle violation fixed. somehow missed the test-patch

oozie/src/main/java/org/apache/falcon/oozie/DatabaseImportWorkflowBuilder.java
vramachan: checkstyle violation fixed. somehow missed the test-patch

oozie/src/main/java/org/apache/falcon/oozie/ImportExportCommon.java
vramachan: all other options are added that way (see below) in the XML. 
                    <argument name="--split-by" value="id"/>
                    <argument name="--num-mappers" value="2"/>
                    <argument name="--direct" value="true"/>
                    <argument name="--direct.default-character-set" value="latin1"/>
                    <argument name="--verbose" value="true"/>

examples/entity/importexport/mysql-feed-import-direct-filesystem.xml
```diff
@@ -0,0 +1,56 @@
...
+                             <field>id</field>
+                             <field>name</field>
+                           </includes>
+                    </fields>
+                </source>
+                <arguments>
+                    <argument name="--split-by" value="id"/>
+                    <argument name="--num-mappers" value="2"/>
+                    <argument name="--direct" value="true"/>
+                    <argument name="--direct.default-character-set" value="latin1"/>
```
vrangan: Let us have a single extraArgs as a string list
For example
   <argument name="ConnectorExtraArg"  value="--charset=utf8 --schema=abc"/>

vramachan: connectorExtaArgs added.

oozie/src/main/java/org/apache/falcon/oozie/ImportExportCommon.java
PraveenAdlakha: sure makes sense..

PR #235, opened 2016-07-20
FALCON-2081 ExtensionManagerIT fails occassionally

bvellanki:
Fixed a rat-check error as part of this Jira as it was blocking testing and it was only 1 line change


vrangan:
+1  Thanks @bvellanki

sowmyaramesh:
+1, thanks for working on this!

PR #236, opened 2016-07-21
Falcon 2093 - Support TD connector for Database Import and Export

vramachan:


peeyushb:
+1. Patch looks good.

yzheng-hortonworks:
+1. Could you add a JIRA for unit tests? Thank you, @vramachan

vramachan:
@yzheng-hortonworks this is specific to using TD connectors. We don't have infrastructure to invoke TD connector during the UT and IT.

bvellanki:
Will commit

PR #237, opened 2016-07-21
FALCON-2090 HDFS Snapshot failed with UnknownHostException when sched…

bvellanki:
…uling in HA Mode


peeyushb:
+1 . Patch looks good.

PR #238, opened 2016-07-22
FALCON-2082 Add CSRF filter for REST APIs

yzheng-hortonworks:
Provide the user an option to enable CSRF filter and configure custom header and browser agents. Forked RestCsrfPreventionFilter from Hadoop 2.8.0 and fixed checkstyle errors. Tested with CSRF filter enabled and custom header "FALCON-CSRF-FILTER" that 1) GET methods always work; 2) POST/DELETE methods will be accepted only if the user provides the correct custom header (See picture below); 3) Falcon CLI works well.

screen shot 2016-07-27 at 4 17 32 pm

screen shot 2016-07-27 at 4 17 07 pm


vrangan:
+1

ajayyadava:
Instead of creating a dependency on Hadoop 2.8, we can just copy the same filter (it's just one class for code and one class for tests).

ajayyadava:
As I had predicted quite some time back, UI specific code has started creeping into our backend (deja-vu @sriksun ? :stuck_out_tongue: ).  In general, I don't like the idea of UI only concerns creeping in REST services, such code are better handled by UI specific frameworks. We already have several commands just for UI and it keeps growing. 

May be we should just switch to a more robust UI framework instead of giving one solution at a time and building everything on our own.

ajayyadava:
Also, this is not the best way to implement CSRF protection. The most recommended way to implement CSRF is the *Synchronizer (CSRF) Tokens*. Mechanisms like above are not safe against plugin vulnerabilities like [this ](http://www.securityfocus.com/archive/1/441014/30/0/threaded )

common/src/main/java/org/apache/falcon/security/SecurityUtil.java
```diff
@@ -115,6 +115,16 @@ public static boolean isAuthorizationEnabled() {
                 "falcon.security.authorization.enabled", "false"));
     }
 
+    /**
+     * Checks if CSRF filter is enabled in the configuration.
+     *
+     * @return true if falcon.security.csrf.enabled is enabled, false otherwise
+     */
+    public static boolean isCSRFFilterEnabled() {
```
ajayyadava: This should also check if the security is enabled as without that there is no authentication.

common/src/main/resources/startup.properties
```diff
@@ -243,6 +243,9 @@ it.workflow.execution.listeners=org.apache.falcon.catalog.CatalogPartitionHandle
 # Authorization Enabled flag: false (default)|true
 *.falcon.security.authorization.enabled=false
 
+# CSRF filter enabled flag: false (default) | true
+*.falcon.security.csrf.enabled=false
```
ajayyadava: It will be a good idea to mention the dependency on other properties in order to enable this property.

prism/src/main/webapp/WEB-INF/web.xml
```diff
@@ -41,6 +41,15 @@
     </filter>
 
     <filter>
+        <filter-name>csrf</filter-name>
+        <filter-class>org.apache.falcon.security.FalconCSRFFilter</filter-class>
+        <init-param>
+            <param-name>methods-to-ignore</param-name>
+            <param-value>GET</param-value>
```
ajayyadava: Should we add HEAD and OPTIONS as well?

prism/src/test/java/org/apache/falcon/security/FalconCSRFFilterTest.java
```diff
@@ -0,0 +1,134 @@
...
+        MockitoAnnotations.initMocks(this);
+    }
+
+    @Test
+    public void testCSRFEnabledAllowedMethodFromBrowser() throws Exception {
+        StartupProperties.get().setProperty("falcon.security.csrf.enabled", "true");
+        mockHeader("Mozilla/5.0", null);
+        mockGetMethod();
+        mockRunFilter();
+        Mockito.verify(mockResponse, Mockito.never()).sendError(HttpServletResponse.SC_BAD_REQUEST, CSRF_ERROR_MESSAGE);
```
ajayyadava: Wouldn't SC_FORBIDDEN be more appropriate?

ajayyadava:
May be I overlooked something obvious, but how is the filter distinguishing between requests from UI / other API calls?

prism/src/test/java/org/apache/falcon/security/FalconCSRFFilterTest.java
```diff
@@ -0,0 +1,134 @@
...
+    @Mock
+    private FilterChain mockChain;
+
+    @Mock
+    private FilterConfig mockConfig;
+
+    @BeforeClass
+    public void setUp() throws Exception {
+        MockitoAnnotations.initMocks(this);
+    }
```
ajayyadava: It will be useful to add tests for pure API clients as well.

vrangan:
>> As I had predicted quite some time back, UI specific code has started creeping into our backend (deja-vu @sriksun ? 😛 ). In general, I don't like the idea of UI only concerns creeping in REST services, such code are better handled by UI specific frameworks.

Not sure how a UI specific framework alone be useful in this case for not adding this filter - There is nothing that prevents a user to have a UI that call the REST api without going through your UI framework

prism/src/main/webapp/WEB-INF/web.xml
yzheng-hortonworks: Good idea. Yes, we can customize the header (e.g. add "falcon-" to header). Header, methods and user agents are the options provided in RestCsrfPreventionFilter currently.

prism/src/test/java/org/apache/falcon/security/FalconCSRFFilterTest.java
yzheng-hortonworks: This depends on the implementation of RestCsrfPreventionFilter. They send SC_BAD_REQUEST if they detect suspected CSRF.

prism/src/test/java/org/apache/falcon/security/FalconCSRFFilterTest.java
yzheng-hortonworks: It is covered in testCSRFEnabledAllowNonBrowserInteractionWithoutHeader. Let me know if this answers your comment.

common/src/main/java/org/apache/falcon/security/SecurityUtil.java
yzheng-hortonworks: Good question. Somehow the current Hadoop CSRF filter RestCsrfPreventionFilter (from their source code) doesn't need kerberos authentication. It provides the standard header check. It also works with our current simple authentication option (i.e. pass user name in url).

common/src/main/resources/startup.properties
yzheng-hortonworks: The first question is if we have dependency on authorization options? The current filter actually works with both simple and kerberos authentication we provided in Falcon. See more comments above.

yzheng-hortonworks:
>> May be I overlooked something obvious, but how is the filter distinguishing between requests from UI / other API calls?
In RestCsrfPreventionFilter, they only perform CSRF check if the request is from the browser.

>> Instead of creating a dependency on Hadoop 2.8, we can just copy the same filter (it's just one class for code and one class for tests). I would prefer to avoid dependency on Hadoop 2.8
Thanks for bringing this up, @ajayyadava . @vrangan and I had a discussion on this before. Because RestCsrfPreventionFilter can be shared among all Hadoop components, including hive, oozie, hbase, etc., we want to leave the implementation and improvements to this class, instead of reinventing the wheel.

>> Also, this is not the best way to implement CSRF protection. The most recommended way to implement CSRF is the Synchronizer (CSRF) Tokens. Mechanisms like above are not safe against plugin vulnerabilities like this
Yes, adding synchronizer tokens would definitely make the filter stronger. @vrangan: Do you know if Hadoop team has this plan to improve RestCsrfPreventionFilter? Thank you.

common/src/main/java/org/apache/falcon/security/SecurityUtil.java
ajayyadava: In the UI, using default configurations, authentication doesn't happen and you can enter any username and password and emulate that user. In that case one doesn't need to do csrf, right?

prism/src/main/webapp/WEB-INF/web.xml
ajayyadava: I was actually suggesting to add HEAD and OPTIONS in default configuration.

ajayyadava:
> Not sure how a UI specific framework alone be useful in this case for not adding this filter - There is nothing that prevents a user to have a UI that call the REST api without going through your UI framework

If user is using his own UI framework, then security concerns of UI are also his concern and not of the REST API. It will be same for other UI concerns like cookie management etc. They make sense only for UI client and should be part of UI and not REST API. The reason that we have to put them in REST API is because our UI can not handle it on its own.

ajayyadava:
>In RestCsrfPreventionFilter, they only perform CSRF check if the request is from the browser.
From the hadoop patch it seems to me that once this is enabled, it will act on all clients (including cli) and not just UI. My question is how is it being determined that a request is from the browser and not through other client say cli?

ajayyadava:
>Because RestCsrfPreventionFilter can be shared among all Hadoop components, including hive, oozie, hbase, etc., we think it is better to leave the implementation and improvements to this class, instead of reinventing the wheel. 

If we were already on Hadoop 2.8 then I would have also agreed. However, currently I am of the opinion that is better to "copy the wheel" and allow existing falcon users to use this feature without requiring them to upgrade to next version of Hadoop.

prism/src/test/java/org/apache/falcon/security/FalconCSRFFilterTest.java
ajayyadava: That is incorrect IMO and another reason to probably "copy the wheel" and fix it on our end rather than waiting for next release of Hadoop.

yzheng-hortonworks:
1. If you looked at the code of RestCsrfPreventionFilter line 112-119, they only filter if the request comes from a browser. We can also configure the browser types that need the filter (I'll add this option to the patch in addition to the customized header as I mentioned in my previous reply). I also tested that Falcon CLI works well with this filter.
{code}
    public void handleHttpInteraction(RestCsrfPreventionFilter.HttpInteraction httpInteraction) throws IOException, ServletException {
        if(this.isBrowser(httpInteraction.getHeader("User-Agent")) && !this.methodsToIgnore.contains(httpInteraction.getMethod()) && httpInteraction.getHeader(this.headerName) == null) {
            httpInteraction.sendError(400, "Missing Required Header for CSRF Vulnerability Protection");
        } else {
            httpInteraction.proceed();
        }

    }
{code}

2. Good point. I'll fork the existing filter to Falcon so we avoid Hadoop upgrade issues. In the future, we need to manually sync up with their improvements if any.

prism/src/main/webapp/WEB-INF/web.xml
yzheng-hortonworks: What do you mean by default configuration? Do you mean startup.properties? That is what I'm going to add to this patch. Thanks, @ajayyadava .

common/src/main/java/org/apache/falcon/security/SecurityUtil.java
yzheng-hortonworks: In this case, it is very possible that CSRF happens by intercepting the user name and making requests with it. This is one type of CSRF we want to prevent. And so yes, we need CSRF in this case.

ajayyadava:
@yzheng-hortonworks Yes that is what I was expecting to see in the patch but I couldn't find. I was looking only at the issue that you provided in the JIRA HADOOP-12691 and that doesn't contain this code. Turns out, it is part of the related JIRA HADOOP-12758. Thanks I will review again.

yzheng-hortonworks:
Ok. Thank you @ajayyadava . I have forked RestCsrfPreventionFilter from Hadoop 2.8.0 and also improved code structure. Going to add options in startup.properties for custom header and browser user agent.

yzheng-hortonworks:
Added configuration properties to allow custom header and browser agent. Also fixed checkstyle errors in Hadoop 2.8.0 RestCsrfPreventionFilter. Tested with CSRF filter enabled and custom header "FALCON-CSRF-FILTER" that 1) GET methods always work; 2) POST/DELETE methods will be accepted only if the user provides the correct custom header. See picture below; 3) Falcon CLI works well. @vrangan @ajayyadava  Could you review? Thank you!

vrangan:
I think the option to integrate the Hadoop 2.8 code into the filter here with the config options is good to provide the basic CsrfProtection wihtout having hadoop 2.8 dependencies.  

+1 for the options also

ajayyadava:
I am away from my system for another 20 days and won't be able to review it. From the comments it seems all my concerns are addressed. 
Request any one of the committers to review and commit it.

sowmyaramesh:
+1. LGTM!

PR #239, opened 2016-07-25
FALCON-2097. Feed update with replication delay creates holes

sandeepSamudrala:


oozie/src/main/java/org/apache/falcon/workflow/engine/OozieWorkflowEngine.java
```diff
@@ -1393,7 +1393,7 @@ private void updateCoords(String cluster, BundleJob bundle,
 
             //calculate next start time based on delay.
             endTime = (delay == null) ? endTime
-                    : EntityUtil.getNextStartTime(coord.getStartTime(), delay, EntityUtil.getTimeZone(entity), endTime);
+                    : EntityUtil.getNextInstanceTimeWithDelay(endTime, delay, EntityUtil.getTimeZone(entity));
```
pallavi-rao: This change basically makes sure that the delay is added to endTime (clusterDef.getValidity().getEnd()). Can you please explain how it solves for the holes, in between, during update?

sandeepSamudrala: EffectiveTime is being sent for update. EndTime is sent only in case of a bundle update where it is supposed to be end time as the one mentioned in the entity cluster's validity.

pallavi-rao: Thanks for clarifying @sandeepSamudrala. Can you please add UTs for the new method? I understand, ITs may be hard for the whole scenario.

pallavi-rao:
The build fails with a checkstyle issue not related to this patch.

Also, would have preferred a few more UT cases across day and month boundaries. But, it is fine.

+1. Will merge shortly.

PR #240, opened 2016-07-25
Falcon 2095: HiveDR fails in NN-HA config

vrangan:
Fix validated in a two cluster configuration with NN/RM HA setup


bvellanki:
+1 for the fix. The jenkins failures are unrelated. So will commit the patch soon.

PR #244, opened 2016-07-27
Falcon-2039 Move falcon post processing to falcon server and remove post processing

PraveenAdlakha:


common/src/main/resources/startup.properties
```diff
@@ -332,3 +333,6 @@ it.workflow.execution.listeners=org.apache.falcon.catalog.CatalogPartitionHandle
 #*.falcon.graphite.port=2003
 #*.falcon.graphite.frequency=1
 #*.falcon.graphite.prefix=falcon
+
+# Property to remove postProcessing
+*.falcon.postprocessing.enable=true
```
pallavi-rao: This doesn't need to be startup property. It can be a runtime property that will allow the admin to change it at runtime.

oozie/src/main/java/org/apache/falcon/oozie/feed/FSReplicationWorkflowBuilder.java
```diff
@@ -58,20 +58,25 @@ public FSReplicationWorkflowBuilder(Feed entity) {
...
         enableCounters(replication);
-        addTransition(replication, SUCCESS_POSTPROCESS_ACTION_NAME, FAIL_POSTPROCESS_ACTION_NAME);
-        workflow.getDecisionOrForkOrJoin().add(replication);
 
-        //Add post-processing actions
-        ACTION success = getSuccessPostProcessAction();
-        addHDFSServersConfig(success, src, target);
-        addTransition(success, OK_ACTION_NAME, FAIL_ACTION_NAME);
-        workflow.getDecisionOrForkOrJoin().add(success);
+        if (!Boolean.parseBoolean(ENABLE_POSTPROCESSING)){
```
pallavi-rao: Can we add a method in the super class and use it across FSReplicationWorkflowBuilder, FeedRetentionWorkflowBuilder and ProcessExecutionWorkflowBuilder?

oozie/src/main/java/org/apache/falcon/service/LogMoverService.java
```diff
@@ -0,0 +1,60 @@
...
+    @Override
+    public void onSuccess(WorkflowExecutionContext context) throws FalconException{
+        if (!Boolean.parseBoolean(ENABLE_POSTPROCESSING)){
+            new JobLogMover().moveLog(context);
+        }
+    }
+
+    @Override
+    public void onFailure(WorkflowExecutionContext context) throws FalconException{
+        onSuccess(context);
```
pallavi-rao: Seems very odd to call onSuccess method in onFailure. May be have a method called onEnd that is called by onSuccess and onFailure

pallavi-rao:
I know you plan to open a separate documentation JIRA. Please also open one to disable post-processing for Import/Export, if you are not comfortable making the changes.

Had challenges while adding UTs?

common/src/main/resources/startup.properties
PraveenAdlakha: Hi ,

This should be a startup.properties for couple of reasons like : 

We generate workflow.xml when the process is submitted or updated so if we keep on playing with runtime.properties user might be under the impression that next instance will not have post processing,if its a runtime properties how will we ensure the behavior across the two instance run.
 
Also if we change the runtime.properties we will change it for other users too who might not be expecting it to happen.

oozie/src/main/java/org/apache/falcon/service/LogMoverService.java
PraveenAdlakha: there are couple of more issue that needs to be addressed here like : 

Its a single threaded it might not scale for minutely jobs and we might end up having a backlog.
Now we are running it through falcon user so might end up having permission issues need to investigate more.Did not had any issues in dev testing as all the users are same in my environment but it will not be the case in production.

I will work on addressing these issues too.Let me know if you have any thoughts.

PraveenAdlakha:
I am comfortable in making the changes for HCat ,Import and Export will make those too...

No challenges for UT will add them too in the current classes.

common/src/main/resources/startup.properties
pallavi-rao: Well, runtime properties are system level settings and will affect all users (only admins make this change). So, one user can't play around with it.

I understand the behavior change that users will notice, but, that is expected and holds for all other properties in runtime.properties too. For example, *.feed.retry.attempts, users will see a change in number of retries based on the value here.

I prefer it being in runtime.properties for the simple reason that this is not something required during startup and an admin can modify it without having to restart the server. Having said that, I don't want to be hung upon where it should go.

oozie/src/main/java/org/apache/falcon/service/LogMoverService.java
pallavi-rao: Good idea on making this multi-threaded.

Regarding the permission, as long as the job logs are readable by world it should be fine, since we copy logs. It will be a problem only when users have explicitly set mapreduce.job.acl-view-job to allow only a specific set of users, in this case, falcon will have to doAs that user.

Btw, the JobLogMover is currently broken. Please file another JIRA to fix it.

PraveenAdlakha:
Hi All ,

I have made the changes for Hcat,Import,Export workflow builder but havenot done any dev testing anyone who is using them can do the dev testing at there end.Rest of the work is dev tested and UT's have been added.

Thanks

oozie/src/main/java/org/apache/falcon/service/LogMoverService.java
```diff
@@ -0,0 +1,112 @@
...
+ */
+public class LogMoverService implements WorkflowExecutionListener  {
+
+    private static final Logger LOG = LoggerFactory.getLogger(LogMoverService.class);
+
+    public static final String ENABLE_POSTPROCESSING = StartupProperties.get().
+            getProperty("falcon.postprocessing.enable");
+
+    private BlockingQueue<Runnable> blockingQueue = new ArrayBlockingQueue<>(50);
+    private ExecutorService executorService = new ThreadPoolExecutor(20, getThreadCount(), 120,
```
pallavi-rao: You could just use `Executors.newFixedThreadPool(numThreads);`
This uses LinkedBlockingQueue and you don't have to handle the waiting logic as you have done in the onEnd method.

oozie/src/main/java/org/apache/falcon/oozie/feed/FeedRetentionWorkflowBuilder.java
```diff
@@ -51,20 +51,26 @@ public FeedRetentionWorkflowBuilder(Feed entity) {
...
 
-        //Add post-processing actions
-        ACTION success = getSuccessPostProcessAction();
-        addTransition(success, OK_ACTION_NAME, FAIL_ACTION_NAME);
-        workflow.getDecisionOrForkOrJoin().add(success);
 
-        ACTION fail = getFailPostProcessAction();
-        addTransition(fail, FAIL_ACTION_NAME, FAIL_ACTION_NAME);
-        workflow.getDecisionOrForkOrJoin().add(fail);
+        if (!isPostProcessingEnabled()){
```
pallavi-rao: Re-iterate. Is it not possible to move this piece of code to a parent class and not repeat it across sub-classes?

oozie/src/test/java/org/apache/falcon/oozie/feed/OozieFeedWorkflowBuilderTest.java
```diff
@@ -261,6 +266,54 @@ private void buildCoordAndValidateFrequency(final String expectedFrequency) thro
...
+        }
+        assertTrue(userAction);
+        assertTrue(postProcessing);
+
+        workflow = getWorkflowapp(trgMiniDFS.getFileSystem(), coord);
+
+        userAction = false;
+        postProcessing = true;
+
+        for(Object action : workflow.getDecisionOrForkOrJoin()){
```
pallavi-rao: You can iterate over the coords in a loop without having to repeat this code. You can look for either eviction or replication inside.

oozie/src/test/java/org/apache/falcon/oozie/feed/OozieFeedWorkflowBuilderTest.java
```diff
@@ -261,6 +266,54 @@ private void buildCoordAndValidateFrequency(final String expectedFrequency) thro
...
+        OozieEntityBuilder builder = OozieEntityBuilder.get(postFeed);
+        Path bundlePath = new Path("/projects/falcon/");
+        builder.build(trgCluster, bundlePath);
+        BUNDLEAPP bundle = getBundle(trgMiniDFS.getFileSystem(), bundlePath);
+        List<COORDINATOR> coords = bundle.getCoordinator();
+        COORDINATORAPP coord = getCoordinator(trgMiniDFS, coords.get(0).getAppPath());
+
+        WORKFLOWAPP workflow = getWorkflowapp(trgMiniDFS.getFileSystem(), coord);
+
+        Boolean userAction = false;
```
pallavi-rao: Nit : Just for better readability, you can call these foundUserAction and foundPostProcessing. Initialize both to false and then assertTrue(foundUserAction) and assertFalse(foundPostProcessing).

oozie/src/test/java/org/apache/falcon/oozie/process/OozieProcessWorkflowBuilderTest.java
```diff
@@ -114,6 +116,8 @@ public void setUp() throws Exception {
         storeEntity(EntityType.FEED, "clicksummary", FEED_XML);
         storeEntity(EntityType.PROCESS, "clicksummary", PROCESS_XML);
         storeEntity(EntityType.PROCESS, "pig-process", PIG_PROCESS_XML);
+        storeEntity(EntityType.PROCESS, "post-process", POST_PROCEES_XML);
```
pallavi-rao: Why create one more process and not just reuse PIG_PROCESS_XML?

oozie/src/test/java/org/apache/falcon/oozie/feed/OozieFeedWorkflowBuilderTest.java
```diff
@@ -99,6 +102,7 @@
     private static final String SRC_CLUSTER_PATH = "/feed/src-cluster.xml";
     private static final String TRG_CLUSTER_PATH = "/feed/trg-cluster.xml";
     private static final String FEED = "/feed/feed.xml";
+    private static final String POST_FEED = "/feed/fs-post-processing-feed.xml";
```
pallavi-rao: Why create one more test file? FEED has both retention and replication that can be used in your test case, right?

PraveenAdlakha: reason I created a new feed because not all the test were calling builder.build function which creates the new workflow everytime. But seems like they donot need it.Will remove the feed and the extra process added for testing.

oozie/src/main/java/org/apache/falcon/service/LogMoverService.java
PraveenAdlakha: We ran in GC while using newFixedThreadPool as LinkedBlockingQueue can grow infinite and we can further extend the code to use RejectionHandler also.It was intentional as we would like to know how much capacity is needed actually.

We can further extend the logic by making the queue size configurable let me know your thoughts on that.

oozie/src/main/java/org/apache/falcon/oozie/feed/FeedRetentionWorkflowBuilder.java
PraveenAdlakha: Sorry from the last comment I understood I have to move the Boolean check to the parent class,which I did.

addTransition method is already in the parent class , can you please be bit more specific weather I have to move whole if/else condition to parent class or only if or something else.

pallavi-rao: Oh.. sorry.. I meant the whole if-else block.

oozie/src/main/java/org/apache/falcon/oozie/OozieOrchestrationWorkflowBuilder.java
```diff
@@ -219,6 +230,25 @@ protected void decorateWorkflow(WORKFLOWAPP wf, String name, String startAction)
         wf.getDecisionOrForkOrJoin().add(kill);
     }
 
+    protected void addPostProcessing(WORKFLOWAPP workflow, ACTION action) throws FalconException{
```
pallavi-rao: This is missing addHDFSServersConfig() method, which was added for replication workflows. But, I don't think it is needed for post-processing (should be sufficient for replication). Can you check @sowmyaramesh or @vramachan to confirm

pallavi-rao:
+1 apart from the point on HDFSServerConfigs.

docs/src/site/twiki/Configuration.twiki
```diff
@@ -317,6 +317,20 @@ su - $OOZIE_USER
...
+---+++Disabling Falcon Post Processing
+Falcon post processing performs two tasks:
+They send user notifications to Active mq.
+It moves oozie executor logs once the workflow finishes.
+
+If post processing is failing because of any reason user mind end up having a backlog in the pipeline thats why it has been made optional.
+
+To disable post processing set the following property to false in startup.properties :
+<verbatim>
+*.falcon.postprocessing.enable=false
```
vrangan: should it be enabled?   This will have changes to method names also.

oozie/src/main/java/org/apache/falcon/oozie/OozieOrchestrationWorkflowBuilder.java
vrangan: As Falcon has the security context, you don't need to add HDFS server configs for post processing.

pallavi-rao: Thanks @vrangan for confirming.

docs/src/site/twiki/Configuration.twiki
PraveenAdlakha: By default it is enabled in startup.properties.In the twiki we are talking about disabling it.

pallavi-rao: @vrangan. Heard you are away for a few days. Will merge this PR for now. Please open a JIRA to follow up on your comment.

pallavi-rao:
@PraveenAdlakha, looks there are code merge conflicts. Please resolve them. Will merge once the conflicts are resolved and build is successful.

PraveenAdlakha:
@pallavi-rao its mergable now.

PR #245, opened 2016-07-28
FALCON-2101 Disable IT tests that depend on instance state

bvellanki:
Unblocking release by disabling indeterministic tests. Create FALCON-2102 for a more thorough investigation and fix.


peeyushb:
+1.

vrangan:
+1

PR #246, opened 2016-07-28
Falcon 2096. Feed instance listing doesn't give instance status of all the clusters

sandeepSamudrala:


sandeepSamudrala:
This pull request is above https://github.com/apache/falcon/pull/239
Don't merge this before the above one gets merged.

common/src/main/java/org/apache/falcon/entity/FeedHelper.java
```diff
@@ -818,9 +819,11 @@ public static FeedInstanceResult getFeedInstanceListing(Entity entityObject,
                 instance.size = feedStatus.getSize();
                 instance.sizeH = feedStatus.getSizeH();
                 instances[index++] = instance;
+                allInstances.add(instance);
```
peeyushb: It looks like that instances[index++] and allInstances.add() both are adding the feed instance. I think instances[] could have been used instead of creating new List variable allInstances by moving FeedInstanceResult.Instance[] instances above clusters for loop.

common/src/main/java/org/apache/falcon/entity/FeedHelper.java
```diff
@@ -818,9 +819,11 @@ public static FeedInstanceResult getFeedInstanceListing(Entity entityObject,
...
             }
-            result.setInstances(instances);
```
PraveenAdlakha: After this change there is no usage for variable instances its redundant...

PR #252, opened 2016-08-01
FALCON-2047 Use /apps/falcon/extensions/mirroring as base dir for HiveDR instead of /apps/data-mirroring/

sowmyaramesh:


vrangan:
+1

src/main/assemblies/distributed-package.xml
```diff
@@ -248,6 +248,16 @@
         </fileSet>
 
         <fileSet>
+            <directory>../</directory>
+            <outputDirectory>extensions/mirroring</outputDirectory>
+            <excludes>
+                <exclude>*/**</exclude>
+            </excludes>
+            <fileMode>0770</fileMode>
```
PraveenAdlakha: Can you please help me understand the reason for extra write permission here..

sowmyaramesh: @PraveenAdlakha : This has been the behavior since HiveDR feature was implemented.

Here is the reason why we falcon expects 770 perm for /apps/data-mirroring now /apps/extensions/mirroring directory :
Directory is used by multiple mirroring jobs to store replication specific data. Deleting this data, even accidentally, can cause mirror jobs to redo already replicated events and this can be expensive. Hence we decided that the directory has to be read+writable by Falcon and multiple users who belong to specific group, which in this case is “users"   This way, rest of the world does not have permission to read/delete the data.

bvellanki:
+1 for the review @sowmyaramesh

peeyushb:
Looks good. +1 .

src/main/assemblies/distributed-package.xml
PraveenAdlakha: thanks for the explanation @sowmyaramesh

PraveenAdlakha:
+1 LGTM

PR #253, opened 2016-08-02
FALCON-2104 Loss of data in GraphDB when upgrading Falcon 0.9 to 0.10

bvellanki:


bvellanki:
This patch has been tested end2end to gaurantee data integrity in graphDB when upgrading from 0.9 to 0.10

vrangan:
+1.

vrangan:
One minor nit - we have many different scripting solutions in the product.   A few python and shell scripts.   May be we add a dependency on Jython and convert all scripts to Python scripts compatible with Jython also and can have consistency and ease.

We currently do package groovy but given lots of existing python files, it maybe faster to just use Jython dependency as a core dependency (and use the version used in Pig which has been vetted for any security vulnerabilities known thus far).

Will start a Discuss thread on this

sowmyaramesh:
+1

PR #254, opened 2016-08-02
FALCON-2107 NPE in FalconWorkflowEngine::isActive() method

bvellanki:


scheduler/src/main/java/org/apache/falcon/workflow/engine/FalconWorkflowEngine.java
```diff
@@ -108,12 +108,17 @@ public void dryRun(Entity entity, String clusterName, Boolean skipDryRun) throws
...
-        if (STATE_STORE.entityExists(id)) {
-            return STATE_STORE.getEntity(id).getCurrentState() != EntityState.STATE.SUBMITTED;
+        try {
+            EntityID id = new EntityID(entity);
+            // Ideally state store should have all entities, but, check anyway.
+            if (STATE_STORE.entityExists(id)) {
+                return STATE_STORE.getEntity(id).getCurrentState() != EntityState.STATE.SUBMITTED;
+            }
+            return false;
+        } catch (NullPointerException npe) {
```
vrangan: May be we should do a refactoring of this for better handling of the case when JPAService is not there in a future patch.

+1 for the current one

bvellanki: Agreed @pallavi-rao  and I talked about this. But the changes will be too disruptive for 0.10 release.

PR #262, opened 2016-08-12
FALCON-2115 UT test failure on FalconCSRFFilterTest

yzheng-hortonworks:
Need to add the property falcon.security.csrf.header to startup properties when testing custom header for CSRF filter.


sowmyaramesh:
+1, will commit the change.

PR #263, opened 2016-08-13
FALCON-1776 Process instance update in titan DB based on JMS notifications on coordinator actions

yzheng-hortonworks:
In JMS notifications from the coordinator action, it is not guaranteed to provide information on timestamp and user workflow version. Need to allow these properties to be optional when adding process instance to titan DB. Also need to address the different format when there is no input/output feed. See the picture below for an example result (i.e. WAITING instance).

screen shot 2016-08-12 at 4 50 59 pm

### DIFF
--- a/common/src/main/java/org/apache/falcon/metadata/InstanceRelationshipGraphBuilder.java
+++ b/common/src/main/java/org/apache/falcon/metadata/InstanceRelationshipGraphBuilder.java
@@ -76,8 +76,8 @@ public class InstanceRelationshipGraphBuilder extends RelationshipGraphBuilder {
         String processInstanceName = getProcessInstanceName(context);
         LOG.info("Adding process instance: {}", processInstanceName);
 
-        Vertex processInstance = addVertex(processInstanceName,
-                RelationshipType.PROCESS_INSTANCE, context.getTimeStampAsLong());
+        Vertex processInstance = addVertex(processInstanceName, RelationshipType.PROCESS_INSTANCE,
+                context.hasTimeStamp() ? context.getTimeStampAsLong() : null);
         addWorkflowInstanceProperties(processInstance, context);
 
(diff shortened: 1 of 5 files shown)


common/src/main/java/org/apache/falcon/metadata/InstanceRelationshipGraphBuilder.java
```diff
@@ -76,8 +76,8 @@ public Vertex addProcessInstance(WorkflowExecutionContext context) throws Falcon
         String processInstanceName = getProcessInstanceName(context);
         LOG.info("Adding process instance: {}", processInstanceName);
 
-        Vertex processInstance = addVertex(processInstanceName,
-                RelationshipType.PROCESS_INSTANCE, context.getTimeStampAsLong());
+        Vertex processInstance = addVertex(processInstanceName, RelationshipType.PROCESS_INSTANCE,
+                context.hasTimeStamp() ? context.getTimeStampAsLong() : null);
```
sandeepSamudrala: If the time is not available,  how will the vertex added be useful?

yzheng-hortonworks: This is the status update time, not the nominal time. We always have nominal time and status information, which are the most important.

bvellanki:
+1 , jenkins test failures seem unrelated.

common/src/main/java/org/apache/falcon/metadata/InstanceRelationshipGraphBuilder.java
sandeepSamudrala: Makes sense. Thanks for the clarification.

sandeepSamudrala:
👍

peeyushb:
+1. Looks good.

PR #264, opened 2016-08-13
FALCON-2112 Set property value to set map memory for replication and retention

sandeepSamudrala:


peeyushb:
@sandeepSamudrala Can't parameters for map task memory be set in the Feed entity properties section.

sandeepSamudrala:
@peeyushb : Yes. They can be set. This way gives it to define globally for all feeds and not each feed has to specify it.

peeyushb:
Each feed entity must have it's own calculated requirement of memory for retention and replication job, so I am more of the opinion let each feed entity provide memory requirements by specifying in properties rather setting it globally across for all feeds.

Thoughts please.

peeyushb:
In addition, "oozie.launcher.mapreduce.map.memory.mb" and "mapreduce.map.memory.mb" is not just one memory parameter for Hadoop jobs. There are bunch of other memory parameters as well and those parameters values are set in accordance with each other.

common/src/main/resources/runtime.properties
```diff
@@ -23,8 +23,9 @@
 
 *.falcon.replication.workflow.maxmaps=5
 *.falcon.replication.workflow.mapbandwidth=100
+*.falcon.feed.workflow.mapreduce.map.memory.mb=512
```
pallavi-rao: Please use the same naming convention as the other properties .. falcon.replication.workflow.map.memory.mb.

sandeepSamudrala:
@peeyushb : Even with this patch, user can specify his own requirements in the feed properties and they have precedence over global settings. In an organization running large clusters, it would be difficult to update all the feeds and so this provides leisure to just touch the feeds and the new settings will get kicked.

lifecycle/src/main/resources/action/feed/eviction-action.xml
```diff
@@ -39,6 +39,7 @@
             </property>
```
pallavi-rao: If you are setting the launcher memory for others, better be consistent and set it here too.

sandeepSamudrala:
@peeyushb : On coming to other parameters. Yes I agree, there are bunch of other parameters which work in accordance with the above settings. But coming to feed replication and retention there is no MR job, they are simple java action to delete data and distcp job to copy the data, which won't require many other settings. 

These properties are set only to feeds and not processes.

lifecycle/src/main/resources/action/feed/eviction-action.xml
sandeepSamudrala: Ack. I will put launcher memory here too.

pallavi-rao:
Please also update documentation. 

The build has failed, but, I guess the ITs that have failed are un-related.

pallavi-rao:
+1 from my end. @peeyushb, are all your comments addressed?

peeyushb:
@sandeepSamudrala Just setting this particular memory parameter alone and that too globally for all the Falcon feed entity scheduled MR jobs, many of the times it will not help, if other memory parameters are not in accordance and user has not defined anything in feed entity to override the global settings.

Also, why you think that update functionality for feed entity can't be leveraged to set the relevant memory  through properties for the required and particular feed entity MR jobs. As definitely not all the feed entities in Falcon may require this. 

If we do comparision with Oozie, if any particular workflow has different memory requirement they just specify the value to the required memory parameters for that particular workflow only. 

Moreover, feed replication and retention run as MR jobs.

vrangan:
Hi @sandeepSamudrala  - setting map memory mb alone is not enough.   Since the minimum allocation will be determined by yarn.scheduler.minimum-allocation-mb and yarn.scheduler.maximum-allocation-mb you may indeed get more than 512mb.

Also, even if you changed the minimum allocation to 512m (which should not be the default as you have), then you need to make sure that the mapreduce.map.java.opts has -Xmx set less than that with allowance for the JVM requirements.

sandeepSamudrala:
@vrangan : Thanks for pointing the same. I had realized what other effects it could introduce. I had made the changes accordingly for the same to accommodate 
yarn.app.mapreduce.am.resource.mb
yarn.app.mapreduce.am.command-opts

mapreduce.map.memory.mb
mapreduce.map.java.opts  

I will test the updated patch and update the pull request.

sandeepSamudrala:
@vrangan @peeyushb : I have made changes accordingly. Please review.

docs/src/site/twiki/EntitySpecification.twiki
```diff
@@ -416,7 +420,12 @@ waiting for the feed instance, parallel decides the concurrent replication insta
 order decides the execution order for replication instances like FIFO, LIFO and LAST_ONLY.
 DistCp options can be passed as custom properties, which will be propagated to the DistCp tool. "maxMaps" represents
 the maximum number of maps used during replication. "mapBandwidth" represents the bandwidth in MB/s
-used by each mapper during replication. "overwrite" represents overwrite destination during replication.
+used by each mapper during replication.
+"mapMemory" represents the mapreduce map memory in mb to be specified to the respective replication and retention jobs.
```
PraveenAdlakha: nit: just to keep it consistent either start it from the same line or move every property name to a separate line.
Right now mapBandwith is in one line and rest of setting is in form of paragraph.

oozie/src/main/java/org/apache/falcon/oozie/feed/FeedReplicationWorkflowBuilder.java
```diff
@@ -145,6 +161,24 @@ private String getDefaultMapBandwidth() {
         return RuntimeProperties.get().getProperty("falcon.replication.workflow.mapbandwidth", "100");
     }
 
+    private String getDefaultAmMemory() {
```
PraveenAdlakha: It would be better to move these method to OozieOrchestrationWorkflowBuilder and use it from there in FeedRetentionWorkflowBuilder and FeedReplicationWorkflowBuilder rather than doing same thing twice.

docs/src/site/twiki/EntitySpecification.twiki
sandeepSamudrala: Ack,will make the changes accordingly.

oozie/src/main/java/org/apache/falcon/oozie/feed/FeedReplicationWorkflowBuilder.java
sandeepSamudrala: These settings are not meant for other builders like ProcessWorfklowBuilder, might give wrong conscience of being available for other builders. Thats why it has been put into these two builders specifically. In case if we want to achieve the same for process builder too then we can move it into OrchestrationWorkflowBuilder.

sandeepSamudrala:
@peeyushb , @vrangan : Can you please review this?

oozie/src/main/java/org/apache/falcon/oozie/feed/FeedReplicationWorkflowBuilder.java
PraveenAdlakha: Still its better to move it to OrchestrationWorkflowBuilder to avoid repetition of code.We can override the method in  ProcessWorfklowBuilder and throw an exception method not supported if donot want to use them there.

sandeepSamudrala: Its just not ProcessWorkflowBuilder , there are data import, export and hive builders.

PraveenAdlakha: As discussed please remove the method and use it inline in both the places.

PR #265, opened 2016-08-13
FALCON-2113 Falcon retry happens in few cases inspite of a manual kill from the user.  move it to a different state(ignore) and check for that state before retrying to stop it from retrying.

sandeepSamudrala:


common/src/main/java/org/apache/falcon/workflow/engine/AbstractWorkflowEngine.java
```diff
@@ -111,7 +114,8 @@ public abstract InstancesResult getInstanceParams(Entity entity, Date start, Dat
 
     public abstract boolean isNotificationEnabled(String cluster, String jobID) throws FalconException;
 
-    public abstract Boolean isWorkflowKilledByUser(String cluster, String jobId) throws FalconException;
+    public abstract Boolean isWorkflowKilledByUser(String cluster, String jobId,
+                                                   String parentId) throws FalconException;
```
pallavi-rao: Adding a parent Id here makes some assumptions on the hierarchy (coord->wf) which does not hold for native workflow engine. Instead, do a WorkflowInfo to retrieve the parent inside the method.

client/src/main/java/org/apache/falcon/resource/InstancesResult.java
```diff
@@ -34,7 +34,7 @@
      * Workflow status as being set in result object.
      */
     public static enum WorkflowStatus {
-        WAITING, RUNNING, SUSPENDED, KILLED, FAILED, SUCCEEDED, ERROR, SKIPPED, UNDEFINED, READY
+        WAITING, RUNNING, SUSPENDED, KILLED, FAILED, SUCCEEDED, ERROR, SKIPPED, UNDEFINED, READY, KILLEDIGNORED
```
pallavi-rao: Use KILLED_OR_IGNORED instead?

pallavi-rao:
+1

sandeepSamudrala:
Rebased the patch and fixed failures.

PraveenAdlakha:
+1

PR #266, opened 2016-08-15
FALCON-2116 Queue defined in <retention-stage> is not getting used in actual retention job

sandeepSamudrala:


pallavi-rao:
+1. Unrelated build test failures noted.

PR #270, opened 2016-08-23
FALCON-2121 If a feed does not exist on a cluster, process submission throws NPE

pallavi-rao:
After the change:
`bin/falcon entity -submit -type process -file examples/entity/filesystem/pig-process.xml
ERROR: Bad Request;default/Feed in does not exist on cluster local2`


sandeepSamudrala:
validateFeedDefinedForCluster is already available as helper method to validate.

Can you modify that method to get the same functionality?

pallavi-rao:
@sandeepSamudrala, had totally missed the helper method. Used that instead.

sandeepSamudrala:
looks good. 👍

pallavi-rao:
Thanks @sandeepSamudrala for the review. Merging...

PR #277, opened 2016-08-30
FALCON-2133: Disable the LogMover from FeedReplication on secure mode. 

muraliramasami:
LogMover is only supported in case of un-secure mode. It has to be disable in the secure mode. In FeedReplication test cases, LogMover assertions will be done after the functionality tested covered. In this cases, we need to verify the functionality but only disable LogMover.


pragya-mittal:
+1

muraliramasami:
Thanks @pragya-mittal for the review.

PR #283, opened 2016-09-21
FALCON-2146 EntitySlaAlertService runs only once irrespective of the …

PraveenAdlakha:
…configured frequency


pallavi-rao:
+1. Will merge once build completes.

PR #284, opened 2016-09-26
FALCON-2148 Entries not removed from sla list when instance succeeds

PraveenAdlakha:


common/src/main/resources/statestore.properties
```diff
@@ -24,6 +24,10 @@
 ## Falcon currently supports derby, mysql and postgreSQL, change url based on DB.
 #*.falcon.statestore.jdbc.url=jdbc:derby:data/falcon.db;create=true
 
+##Note : In case you are using mysql or postgres please specify URL as
+##       jdbc:mysql://localhost:3306/dbname?serverTimezone=UTC
+##       here serverTimezone should be same a that of oozie server.
```
pallavi-rao: Correct : here serverTimezone should be same a that of oozie server.
to
Here serverTimezone should be same as that of Oozie server.

Same comment applies to change below

pallavi-rao:
+1. Will merge once build finishes.

PR #286, opened 2016-09-26
FALCON-2147 Excess logging due to SlaMonitoring service and Backlog e…

pallavi-rao:
…mitter service

### DIFF
--- a/prism/src/main/java/org/apache/falcon/service/BacklogMetricEmitterService.java
+++ b/prism/src/main/java/org/apache/falcon/service/BacklogMetricEmitterService.java
@@ -145,7 +145,7 @@ public final class BacklogMetricEmitterService implements FalconService,
                 List<MetricInfo> metricsInDB = entry.getValue();
                 List<MetricInfo> metricInfoList = Collections.synchronizedList(metricsInDB);
                 entityBacklogs.put(entry.getKey(), metricInfoList);
-                LOG.debug("Backlog of entity " + entry.getKey().getName() + " for instances " + metricInfoList);
+                LOG.debug("Initializing backlog for entity " + entry.getKey().getName());
             }
         }
     }
@@ -204,7 +204,7 @@ public final class BacklogMetricEmitterService implements FalconService,
 
         @Override
         public void run() {
-            LOG.debug("BacklogMetricEmitter running for entities");
+            LOG.trace("BacklogMetricEmitter running for entities");
             executor = new ScheduledThreadPoolExecutor(10);
             List<Future> futures = new ArrayList<>();
             try {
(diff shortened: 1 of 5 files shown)


prism/src/main/java/org/apache/falcon/service/EntitySLAMonitoringService.java
```diff
@@ -400,7 +399,8 @@ void addNewPendingEntityInstances(Date to, String entityType) throws FalconExcep
                                 currentCluster.getName()).getEnd();
                     }
                     while (nextInstanceTime.before(to) && nextInstanceTime.before(endDate)) {
-                        LOG.info("Adding instance={} for <entity,cluster>={}", nextInstanceTime, key);
+                        LOG.debug("Adding pending instance={} for <entity,cluster>={}; entityType={}",
+                                nextInstanceTime, key, entityType);
```
sandeepSamudrala: Does this print cleanly or should it be changed to key.first and key.second?

PraveenAdlakha: I believe we should change this one to trace as this will print so many lines in the logs.

PraveenAdlakha:
except one comments looks good.

prism/src/main/java/org/apache/falcon/service/BacklogMetricEmitterService.java
```diff
@@ -302,7 +303,7 @@ public void run() {
 
         @Override
         public void run() {
-            LOG.trace("BacklogCheckService running for entities");
+            LOG.debug("BacklogCheckService running for entities");
```
sandeepSamudrala: can you change the message to "BacklogCheckService periodic run started"?

prism/src/main/java/org/apache/falcon/service/EntitySLAMonitoringService.java
```diff
@@ -303,14 +303,14 @@ public void init() throws FalconException {
 
         freq = StartupProperties.get().getProperty("entity.sla.lookAheadWindow.millis", "900000");
         lookAheadWindowMillis = Integer.parseInt(freq);
-        LOG.debug("No old state exists at: {}, Initializing a clean state.", filePath.toString());
+        LOG.info("Initializing EntitySLAMonitoringService from ", filePath.toString());
```
sandeepSamudrala: filePath is no more required and it doesn't initialize anything from the filePath, this line along with the filePath variable can be removed.

pallavi-rao: Agree. However, I still see some dependencies on filePath. We should do a clean remove of it from all places including properties file. Lets handle that in a separate JIRA

pallavi-rao:
Will merge this, if there are no further comments...

PR #290, opened 2016-10-24
FALCON-2171 When feed from multiple colos are replicated, the colo folders get overwritten

pallavi-rao:
Made changes to use the include path to set source and target of Distcp (not just the base directory of feed).

UT added. Manual testing done with single colo. Test on distributed setup pending.


sandeepSamudrala:
:+1:  Test failure is irrelevant to this patch.

PR #292, opened 2016-10-26
FALCON-2175 java.lang.IllegalArgumentException in LogMover service

pallavi-rao:


oozie/src/main/java/org/apache/falcon/logging/JobLogMover.java
```diff
@@ -84,17 +85,23 @@ public void moveLog(WorkflowExecutionContext context){
 
     public int run(WorkflowExecutionContext context) {
         try {
-            OozieClient client = new OozieClient(context.getWorkflowEngineUrl());
+            String engineUrl = context.getWorkflowEngineUrl();
+            if (StringUtils.isBlank(engineUrl)) {
+                LOG.warn("Unable to retrieve workflow url for {} with status {} ",
+                        context.getWorkflowId(), context.getWorkflowStatus());
+                return 0;
```
sandeepSamudrala: there could be a case where the oozie host has been moved from when the instance ran and the current running oozie.
In which case this will not work. Should we not try with the current oozie url? to get the status of the workflow id?

pallavi-rao: If the host has moved, the log movement anyway can't happen as the new host cannot return the old logs.

praveen8927:
looks good +1

PR #293, opened 2016-10-26
FALCON-732  java.lang.IllegalArgumentException in LogMover service

pallavi-rao:


pallavi-rao:
Merging this as it is the same as https://github.com/apache/falcon/pull/292/

PR #298, opened 2016-11-15
FALCON-2183 Extension Builder changes to support new user extensions

sandeepSamudrala:


extensions/src/main/java/org/apache/falcon/extensions/ExtensionBuilder.java
```diff
@@ -19,14 +19,52 @@
...
+     * @throws FalconException
+     */
+    List<Entity> getEntities(final String extensionName, final InputStream extensionStream) throws FalconException;
+
+    /**
+     * @param extensionName extension name.
+     * @return A graph comprising of the whole dag details of the extension.
+     * @throws FalconException
+     */
+    String getExtensionDescription(final String extensionName) throws FalconException;
```
pallavi-rao: The Description can be retrieved by the CLI by just doing a 'cat' on the README. So, we won't need this.

extensions/src/main/java/org/apache/falcon/extensions/ExtensionBuilder.java
```diff
@@ -19,14 +19,52 @@
...
+
+    /**
+     * @param extensionName extension name.
+     * @param extensionProperties Properties supplied will be validated.
+     * @throws FalconException
+     */
+    void validateExtension(final String extensionName, final Properties extensionProperties) throws FalconException;
```
pallavi-rao: The second argument should be an InputStream.

extensions/src/main/java/org/apache/falcon/extensions/ExtensionBuilder.java
```diff
@@ -19,14 +19,52 @@
...
+     * @throws FalconException
+     */
+    List<Pair<String, Schema>> getOutputSchemas(final String extensionName) throws FalconException;
+
+    /**
+     * @param extensionName extension name.
+     * @return Default template for the given extension.
+     * @throws FalconException
+     */
+    String getDefaultTemplate(final String extensionName) throws FalconException;
```
pallavi-rao: Hmm.. This again, we thought we can just do a 'cat' of the template from the CLI. As long the extension packages it (which will be validated during register), CLI can access it.

extensions/src/main/java/org/apache/falcon/extensions/ExtensionBuilder.java
```diff
@@ -19,14 +19,39 @@
...
+     */
+    List<Entity> getEntities(final String extensionName, final InputStream extensionConfigStream)
+        throws FalconException;
+
+    /**
+     * @param extensionName extension name.
+     * @param extensionConfigStream Properties supplied will be validated.
+     * @throws FalconException
+     */
+    void validateExtension(final String extensionName, final InputStream extensionConfigStream) throws FalconException;
```
pallavi-rao: Apologies for not pointing this out earlier, validateExtension is misleading. We are actually validating the configuration supplied by the user. Please change accordingly.

extensions/src/main/java/org/apache/falcon/extensions/Extension.java
```diff
@@ -92,11 +104,21 @@ private static String getWFPath(final Map<String, String> extensionResources,
...
             throw new FalconException("Entity created from the extension template cannot be null");
         }
         LOG.info("Extension processing complete");
-        return Arrays.asList(entity);
+        return Collections.singletonList(entity);
+    }
+
+    @Override
+    public void validateExtension(String extensionName, InputStream extensionConfigStream) throws FalconException {
+        throw new UnsupportedException("Not yet Implemented");
```
pallavi-rao: Apologies again for not pointing out earlier. Why unsupported? This could invoke org.apache.falcon.extensions.Extension#validateProperties. isn't it?

pallavi-rao:
👍 . Looks good. Will merge once the build is fixed.

peeyushb:
Looks good, except one comment one comment on getOutputSchema() .

extensions/src/main/java/org/apache/falcon/extensions/ExtensionBuilder.java
```diff
@@ -19,14 +19,39 @@
...
+     */
+    void validateExtensionConfig(final String extensionName, final InputStream extensionConfigStream)
+        throws FalconException;
+
+    /**
+     * @param extensionName extension name.
+     * @return List of the feed names along with the schema that the extension has generated if any.
+     * @throws FalconException
+     */
+    List<Pair<String, Schema>> getOutputSchemas(final String extensionName) throws FalconException;
```
sandeepSamudrala: It would return whatever the extension has defined for the output feeds, would be empty list if not defined any .

PR #301, opened 2016-11-17
FALCON-2186 Rest api to get location of an extension

PraveenAdlakha:
Dev testing :  

bin/falcon extension -location -extensionName hdfs-mirroring

file:/usr/local/falcon/falcon-0.11-SNAPSHOT/extensions/hdfs-mirroring


pallavi-rao:
@PraveenAdlakha Instead of having an API just to return the location, we can provide a generic API to return all details (whatever we store as a metadata bean). That way it is extensible. If new attributes get added to an extension metadata, everything will get returned, the Client can then filter out what it requires.

PraveenAdlakha:
$ bin/falcon extension -detail -extensionName test 

{"detail":{"name":"test","type":"Custom extension","description":"test ing","location":"hdfs:hostname:8020\/user\/dataqa\/sampleextension"}}

prism/src/main/java/org/apache/falcon/resource/extensions/ExtensionManager.java
```diff
@@ -420,6 +421,22 @@ public String getExtensionDescription(
...
 
+    @GET
+    @Path("detail/{extension-name}")
+    @Produces({MediaType.APPLICATION_JSON})
+    public Response getDetail(@PathParam("extension-name") String extensionName){
+        checkIfExtensionServiceIsEnabled();
+        validateExtensionName(extensionName);
+        JSONObject result = new JSONObject();
+        try {
+            result.put(EXTENSION_DETAIL,buildDetailResult(extensionName));
```
sandeepSamudrala: why JSON inside a JSON? Can you send the json object from buildDetailResult directly?

prism/src/main/java/org/apache/falcon/resource/extensions/ExtensionManager.java
```diff
@@ -524,6 +541,26 @@ private void setEntityTags(Entity entity, String tags) {
...
+        ExtensionMetaStore metaStore = ExtensionStore.get().getMetaStore();
+
+        if (!metaStore.checkIfExtensionExists(extensionName)){
+            throw new StoreAccessException(new Exception("No extension resources found for " + extensionName));
+        }
+
+        ExtensionMetadataBean bean = metaStore.getDetail(extensionName);
+        JSONObject resultObject = new JSONObject();
+        try {
+            resultObject.put(EXTENSION_NAME, bean.getExtensionName().toLowerCase());
```
sandeepSamudrala: why toLowerCase? Can you make changes to Bean to handle for the same so that all consumers are not worried about the same?

prism/src/main/java/org/apache/falcon/resource/extensions/ExtensionManager.java
```diff
@@ -524,6 +541,26 @@ private void setEntityTags(Entity entity, String tags) {
...
+            resultObject.put(EXTENSION_TYPE, bean.getExtensionType());
+            resultObject.put(EXTENSION_DESC, bean.getDescription());
+            resultObject.put(EXTENSION_LOCATION, bean.getLocation());
+        } catch (JSONException e) {
+            throw new FalconException(e);
```
sandeepSamudrala: LOG please

extensions/src/main/java/org/apache/falcon/extensions/store/ExtensionStore.java
```diff
@@ -250,7 +250,7 @@ public String getExtensionResource(final String resourcePath) throws StoreAccess
         }
         return extesnionList;
     }
-
+
```
sandeepSamudrala: Can you remove the tabbing?

cli/src/main/java/org/apache/falcon/cli/FalconExtensionCLI.java
```diff
@@ -158,6 +162,7 @@ public Options createExtensionOptions() {
         Option resume = new Option(FalconCLIConstants.RESUME_OPT, false, "Resume an extension job");
         Option delete = new Option(FalconCLIConstants.DELETE_OPT, false, "Delete an extension job");
         Option unregister = new Option(FalconCLIConstants.UREGISTER, false, "Delete metadata of extension job");
```
pallavi-rao: This can also have a better description. "Un-register an extension. This will make the extension unavailable for instantiation".

cli/src/main/java/org/apache/falcon/cli/FalconExtensionCLI.java
```diff
@@ -158,6 +162,7 @@ public Options createExtensionOptions() {
...
+        Option detail = new Option(FalconCLIConstants.DETAIL, false, "Give location of extension job");
```
pallavi-rao: Description is wrong. "Show details of a given extension"

prism/src/main/java/org/apache/falcon/resource/extensions/ExtensionManager.java
```diff
@@ -524,6 +541,26 @@ private void setEntityTags(Entity entity, String tags) {
         }
     }
 
+    private  JSONObject buildDetailResult(final String extensionName) throws FalconException {
+        ExtensionMetaStore metaStore = ExtensionStore.get().getMetaStore();
+
+        if (!metaStore.checkIfExtensionExists(extensionName)){
+            throw new StoreAccessException(new Exception("No extension resources found for " + extensionName));
```
pallavi-rao: Unnecessary nested exception creation. You can throw ValidationException instead.

prism/src/main/java/org/apache/falcon/resource/extensions/ExtensionManager.java
PraveenAdlakha: its not needed removing it.We are storing the name as it is given by user.

PR #302, opened 2016-11-18
FALCON-2189 Falcon build fix to rearrange startup.properties for extension store now depends on jpa service

sandeepSamudrala:


pallavi-rao:
+1. Will merge once the build goes through.

PR #304, opened 2016-11-19
FALCON-2188 Rest api to register extension

PraveenAdlakha:
$ bin/falcon extension -register -extensionName test -path hdfs://hostname:8020/user/dataqa/sampleextension -description "test ing"

Extension:testregistered succesfully.

Also we get file not found exception if jars or Readme is not present please look at the test cases.

### DIFF
--- a/extensions/src/main/java/org/apache/falcon/extensions/store/ExtensionStore.java
+++ b/extensions/src/main/java/org/apache/falcon/extensions/store/ExtensionStore.java
@@ -275,10 +275,10 @@ public final class ExtensionStore {
         conf.set("fs.default.name", uri.getScheme() + "://" + uri.getAuthority());
         FileSystem fileSystem =  HadoopClientFactory.get().createFalconFileSystem(uri);
         try {
-          fileSystem.listStatus(new Path(uri.getPath() + "/README"));
+            fileSystem.listStatus(new Path(uri.getPath() + "/README"));
         } catch (IOException e){
             LOG.error("Exception in registerExtensionMetadata:", e);
-            throw new ValidationException("README file is not present in the" + uri.getPath());
+            throw new ValidationException("README file is not present in the " + path);
         }
         PathFilter filter=new PathFilter(){
             public boolean accept(Path file){
@@ -302,7 +302,7 @@ public final class ExtensionStore {
         if (!metaStore.checkIfExtensionExists(extensionName)){
             metaStore.storeExtensionMetadataBean(extensionName, path, ExtensionType.CUSTOM, description);
         }else{
-            throw new ValidationException( extensionName + " already exsists.");
+            throw new ValidationException(extensionName + " already exsists.");
         }
         return "Extension :" + extensionName + " registered succesfully.";
     }
--- a/extensions/src/main/java/org/apache/falcon/extensions/store/ExtensionStore.java
+++ b/extensions/src/main/java/org/apache/falcon/extensions/store/ExtensionStore.java
@@ -285,17 +285,26 @@ public final class ExtensionStore {
                 return file.getName().endsWith(".jar");
             }
         };
+        FileStatus[] jarStatus;
         try {
-            fileSystem.listStatus(new Path(uri.getPath() + "/libs/build"), filter);
+            jarStatus = fileSystem.listStatus(new Path(uri.getPath() + "/libs/build"), filter);
+            if (jarStatus.length <=0) {
+                throw new ValidationException("Jars are not present in the " + uri.getPath() + "libs/build.");
+            }
         } catch (IOException e){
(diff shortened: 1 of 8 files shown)


extensions/src/main/java/org/apache/falcon/extensions/store/ExtensionStore.java
```diff
@@ -265,6 +269,28 @@ public String deleteExtensionMetadata(final String extensionName) throws Validat
...
+        URI uri = new URI(path);
+        conf.set("fs.default.name", uri.getScheme() + "://" + uri.getAuthority());
+        FileSystem fileSystem =  HadoopClientFactory.get().createFalconFileSystem(uri);
+        FileStatus[] fileStatusReadMe = fileSystem.listStatus(new Path(uri.getPath() + "/README"));
+        PathFilter filter=new PathFilter(){
+            public boolean accept(Path file){
+                return file.getName().endsWith(".jar");
+            }
+        };
+        FileStatus[] fileStatusJar = fileSystem.listStatus(new Path(uri.getPath() + "/libs"), filter);
```
pallavi-rao: Believe we should check under libs/build.

extensions/src/main/java/org/apache/falcon/extensions/store/ExtensionStore.java
```diff
@@ -265,6 +269,28 @@ public String deleteExtensionMetadata(final String extensionName) throws Validat
         }
     }
 
+    public String registerExtensionMetadata(final String extensionName, final String path, final String description)
+        throws URISyntaxException, FalconException, IOException{
+        Configuration conf = new Configuration();
+        URI uri = new URI(path);
+        conf.set("fs.default.name", uri.getScheme() + "://" + uri.getAuthority());
+        FileSystem fileSystem =  HadoopClientFactory.get().createFalconFileSystem(uri);
+        FileStatus[] fileStatusReadMe = fileSystem.listStatus(new Path(uri.getPath() + "/README"));
```
pallavi-rao: Guess it is sufficient to check for the existence of README and check if it is a file.

extensions/src/main/java/org/apache/falcon/extensions/store/ExtensionStore.java
```diff
@@ -265,6 +269,28 @@ public String deleteExtensionMetadata(final String extensionName) throws Validat
...
+        PathFilter filter=new PathFilter(){
+            public boolean accept(Path file){
+                return file.getName().endsWith(".jar");
+            }
+        };
+        FileStatus[] fileStatusJar = fileSystem.listStatus(new Path(uri.getPath() + "/libs"), filter);
+
+        if (fileStatusReadMe.length == 1 && fileStatusJar.length >0){
```
pallavi-rao: Additional check for existence at least 1 file under META is required.

extensions/src/main/java/org/apache/falcon/extensions/store/ExtensionStore.java
```diff
@@ -265,6 +269,28 @@ public String deleteExtensionMetadata(final String extensionName) throws Validat
...
+            metaStore.storeExtensionMetadataBean(extensionName, path, ExtensionType.CUSTOM, description);
+        }else{
+            throw new ValidationException("Packaging structure is not currect please refer to documentation");
```
pallavi-rao: Error message : Packaging structure of extension ${extensionName} is not correct. Please refer the documentation for details

pallavi-rao: Also, it is preferable to do individual checks and print the appropriate error message, rather than asking user to refer docs.

PraveenAdlakha:
@pallavi-rao  ExtensionService will only be up at prism and that will take care of the requests. Please correct if I am missing some thing here.

extensions/src/main/java/org/apache/falcon/extensions/store/ExtensionStore.java
PraveenAdlakha: Then we need to correct the doc too because in the doc there is no mention of libs/build or libs/runtime.

pallavi-rao:
@PraveenAdlakha ExtensionService may be up only for Prism. However, ExtensionManager is available for prism and falcon webapps. Hence the request will be accepted and may result in an improper error.

PraveenAdlakha:
ohhkk got it..Then we can do what we are doing with EntityManager and InstanceManager for better error handling.Will take it up as a separate jira.

pallavi-rao:
Will merge this. There is a test case that this PR breaks. I will address that with a separate PR as @PraveenAdlakha is out on vacation.

PR #312, opened 2016-12-02
FALCON-2205 describe for non trusted recipe

PraveenAdlakha:


PraveenAdlakha:
Dev Testing:  

$ ./bin/falcon extension -describe  -extensionName test
SLF4J: Class path contains multiple SLF4J bindings.
SLF4J: Found binding in [jar:file:/usr/local/falcon/falcon-0.11-SNAPSHOT/client/lib/slf4j-log4j12-1.7.5.jar!/org/slf4j/impl/StaticLoggerBinder.class]
SLF4J: Found binding in [jar:file:/usr/local/falcon/falcon-0.11-SNAPSHOT/client/lib/falcon-shell-0.11-SNAPSHOT.jar!/org/slf4j/impl/StaticLoggerBinder.class]
SLF4J: See http://www.slf4j.org/codes.html#multiple_bindings for an explanation.
SLF4J: Actual binding is of type [org.slf4j.impl.Log4jLoggerFactory]
test

extensions/src/main/java/org/apache/falcon/extensions/store/ExtensionStore.java
```diff
@@ -140,19 +141,29 @@ private FileSystem initializeFileSystem() {
...
+                fileStatusListIterator = fs.listFiles(extensionPath, true);
+            }else{
+                ExtensionBean extensionBean = metaStore.getDetail(extensionName);
+                extensionPath = new Path(extensionBean.getLocation());
+                FileSystem fileSystem = getHdfsFileSystem(extensionBean.getLocation());
+                fileStatusListIterator = fileSystem.listFiles(extensionPath, true);
+            }
 
             if (!fileStatusListIterator.hasNext()) {
                 throw new StoreAccessException(new Exception(" For extension " + extensionName
```
pallavi-rao: I know this is not part of your patch, but, while you are at it. Can fix this nested exception creation, please?

extensions/src/main/java/org/apache/falcon/extensions/store/ExtensionStore.java
```diff
@@ -218,17 +229,24 @@ public String getExtensionLibPath(final String extensionName) throws StoreAccess
         }
     }
 
-    public String getExtensionResource(final String resourcePath) throws StoreAccessException {
+    public String getExtensionResource(final String resourcePath) throws FalconException {
         if (StringUtils.isBlank(resourcePath)) {
             throw new StoreAccessException(new Exception("Resource path cannot be null or empty"));
```
pallavi-rao: Remove nested exception, while you are at it.

extensions/src/main/java/org/apache/falcon/extensions/store/ExtensionStore.java
```diff
@@ -218,17 +229,24 @@ public String getExtensionLibPath(final String extensionName) throws StoreAccess
         }
     }
 
-    public String getExtensionResource(final String resourcePath) throws StoreAccessException {
+    public String getExtensionResource(final String resourcePath) throws FalconException {
```
pallavi-rao: This method is used only by trusted extension. Custom extensions won't be using this. Hence no change required. However it is good to keep it generic.

extensions/src/main/java/org/apache/falcon/extensions/store/ExtensionStore.java
```diff
@@ -218,17 +229,24 @@ public String getExtensionLibPath(final String extensionName) throws StoreAccess
...
         }
 
         try {
             Path resourceFile = new Path(resourcePath);
+            InputStream data;
 
             ByteArrayOutputStream writer = new ByteArrayOutputStream();
-            InputStream data = fs.open(resourceFile);
-            IOUtils.copyBytes(data, writer, fs.getConf(), true);
+            if (resourcePath.startsWith("hdfs")){
```
pallavi-rao: The scheme could be webhdfs or hftp.. or anything else... A better option would be to check for file:// and assume HDFS in the else part. 

Even better would be to take in the extension name also, see if it is trusted or not and take action appropriately.

extensions/src/main/java/org/apache/falcon/extensions/store/ExtensionStore.java
```diff
@@ -140,19 +141,29 @@ private FileSystem initializeFileSystem() {
...
-    public Map<String, String> getExtensionArtifacts(final String extensionName) throws StoreAccessException {
+    public Map<String, String> getExtensionArtifacts(final String extensionName) throws
+             FalconException {
         Map<String, String> extensionFileMap = new HashMap<>();
+        Path extensionPath;
         try {
-            Path extensionPath = new Path(storePath, extensionName.toLowerCase());
-            RemoteIterator<LocatedFileStatus> fileStatusListIterator = fs.listFiles(extensionPath, true);
+            RemoteIterator<LocatedFileStatus> fileStatusListIterator;
+            if (AbstractExtension.isExtensionTrusted(extensionName)){
```
pallavi-rao: Should we rather not use the ExtensionMetaStore? Better would be to even get rid of AbstractExtension.isExtensionTrusted altogether and use ExtensionMetaStore.

extensions/src/main/java/org/apache/falcon/extensions/store/ExtensionStore.java
```diff
@@ -140,19 +141,29 @@ private FileSystem initializeFileSystem() {
...
+            }
 
             if (!fileStatusListIterator.hasNext()) {
                 throw new StoreAccessException(new Exception(" For extension " + extensionName
                         + " there are no artifacts at the extension store path " + storePath));
             }
             while (fileStatusListIterator.hasNext()) {
                 LocatedFileStatus fileStatus = fileStatusListIterator.next();
-                Path filePath = Path.getPathWithoutSchemeAndAuthority(fileStatus.getPath());
+                Path filePath = fileStatus.getPath();
```
pallavi-rao: without, Path.getPathWithoutSchemeAndAuthority(), this will not work, as in the describe method, only the file name (README) is used for lookup.

extensions/src/main/java/org/apache/falcon/extensions/store/ExtensionStore.java
PraveenAdlakha: I thought about it but it will require a lot of changes around how we bring up ExtensionService as we register trusted extensions at that time.
Which call the same method.Let me know if you are fine with that.

extensions/src/main/java/org/apache/falcon/extensions/store/ExtensionStore.java
PraveenAdlakha: Sorry didnot get you here .

I tested for both trusted and untrusted extensions it was working.

extensions/src/main/java/org/apache/falcon/extensions/store/ExtensionStore.java
PraveenAdlakha: sounds fine.

extensions/src/main/java/org/apache/falcon/extensions/store/ExtensionStore.java
pallavi-rao: I looked at the ExtensionService again:
`List<String> extensions = getExtensions();
            for (String extension : extensions) {`
This list of extensions is only be trusted, so, won't need that check there. isn't it?

extensions/src/main/java/org/apache/falcon/extensions/store/ExtensionStore.java
pallavi-rao: Oh.. Then, may be I'm missing something. Basically, this is my observation:
org.apache.falcon.resource.extensions.ExtensionManager#getExtensionDescription has `ExtensionStore.get().getResource(extensionName, README);`

where README is just "README". Just the file name and not file://... or hdfs://.... , So, as long as fs.listFiles (both for HDFS and localFS) returns path without scheme, it is fine, I guess.

PR #314, opened 2016-12-06
 FALCON-2194 Enhaced validate API to support config validation for user extensions

sandeepSamudrala:


client/src/main/java/org/apache/falcon/ExtensionHandler.java
```diff
@@ -70,7 +76,7 @@
                     + extensionName);
         }
 
-        ExtensionBuilder extensionBuilder = null;
+        ExtensionBuilder extensionBuilder;
```
pallavi-rao: Always better to have initialization. Why remove?

client/src/main/java/org/apache/falcon/ExtensionHandler.java
```diff
@@ -186,4 +192,27 @@ private static String createStagePath(String extensionName, String jobName) {
         urls.add(fileURL);
         return urls;
     }
+
+
+    public static String getExtensionLocation(String extensionName, JSONObject extensionDetailJson) {
```
pallavi-rao: This method has copy-paste issues. Replace all references to type with location.

client/src/main/java/org/apache/falcon/client/FalconClient.java
```diff
@@ -1073,16 +1073,16 @@ private FormDataMultiPart getEntitiesForm(String extensionName, String jobName,
...
 
-        for (Entity entity : entities) {
-            if (EntityType.FEED.equals(entity.getEntityType())) {
-                formDataMultiPart.field("feeds", entity, MediaType.APPLICATION_XML_TYPE);
-            } else if (EntityType.PROCESS.equals(entity.getEntityType())) {
-                formDataMultiPart.field("processes", entity, MediaType.APPLICATION_XML_TYPE);
+        if (entities != null && !entities.isEmpty()) {
+            for (Entity entity : entities) {
+                if (EntityType.FEED.equals(entity.getEntityType())) {
+                    formDataMultiPart.field(FEEDS, entity, MediaType.APPLICATION_XML_TYPE);
```
pallavi-rao: Have you tested this with multiple feeds and processes? From the looks of it, you are overwriting the same fieldName with different entity definition.

client/src/main/java/org/apache/falcon/client/FalconClient.java
```diff
@@ -1154,13 +1152,20 @@ public APIResult updateExtensionJob(final String extensionName, final String fil
...
-                .path(ExtensionOperations.VALIDATE.path, extensionName)
-                .addQueryParam(DO_AS_OPT, doAsUser)
-                .call(ExtensionOperations.VALIDATE, entityStream);
-        return getResponse(APIResult.class, clientResponse);
+    public APIResult validateExtensionJob(final String extensionName, final String jobName,
+                                          final String configPath, final String doAsUser) {
+        String extensionType = ExtensionHandler.getExtensionType(extensionName, getExtensionJobDetail(extensionName));
+        InputStream configStream = getServletInputStream(configPath);
+        if (ExtensionType.TRUSTED.name().equalsIgnoreCase(extensionType)) {
+            ClientResponse clientResponse = new ResourceBuilder()
```
pallavi-rao: Can you update ExtensionManager to return an exception if validate is called on custom extensions?

client/src/main/java/org/apache/falcon/client/FalconClient.java
```diff
@@ -1094,42 +1094,40 @@ private FormDataMultiPart getEntitiesForm(String extensionName, String jobName,
...
-        JSONObject responseJson;
-        try {
-            responseJson = new JSONObject(clientResponse.getEntity(String.class));
-        } catch (JSONException e) {
-            OUT.get().print("Submit failed. Failed to get details for the given extension");
-            throw new FalconCLIException("Submit failed. Failed to get details for the given extension");
-        }
-        String extensionType;
-        String extensionBuildLocation;
+    private JSONObject getExtensionJobDetail(String extensionName) {
```
pallavi-rao: method name should be getExtensionDetail and not getExtensionJobDetail.

PR #316, opened 2016-12-07
FALCON-2190 server side checks to not let any entity operations on entities generated by extensions

sandeepSamudrala:


prism/src/main/java/org/apache/falcon/resource/proxy/SchedulableEntityManagerProxy.java
```diff
@@ -378,6 +385,32 @@ protected APIResult doExecute(String colo) throws FalconException {
...
+                throw FalconWebException.newAPIException("Entity operation is not allowed on this entity as it is"
+                        + "part of an extension job:" + jobName);
+            }
+        }
+    }
+
+    private void isEntityPartOfAnExtension(String type, String entityName) {
+        String tags;
+        try {
+            tags = EntityUtil.getEntity(type, entityName).getTags();
```
pallavi-rao: You can just do a isEntityPartOfAnExtension(EntityUtil.getEntity(type, entityName))

prism/src/main/java/org/apache/falcon/resource/proxy/SchedulableEntityManagerProxy.java
```diff
@@ -378,6 +385,32 @@ protected APIResult doExecute(String colo) throws FalconException {
...
+    private void isEntityPartOfAnExtension(Entity entity) {
+        String tags = entity.getTags();
+        checkExtensionJobExist(tags);
+    }
+
+    private void checkExtensionJobExist(String tags) {
+        if (tags != null) {
+            String jobName = ExtensionManager.getJobNameFromTag(tags);
+            ExtensionMetaStore extensionMetaStore = ExtensionStore.getMetaStore();
+            if (jobName != null && extensionMetaStore.checkIfExtensionJobExists(jobName)) {
```
pallavi-rao: During submit, checkIfExtensionJobExists might return false if that extension job is not submitted yet via the extension API. In such a case, the submit will go through..

sandeepSamudrala: Made the changes to only check for the tag.

prism/src/main/java/org/apache/falcon/resource/proxy/SchedulableEntityManagerProxy.java
```diff
@@ -163,6 +166,7 @@ public APIResult submit(
         Map<String, APIResult> results = new HashMap<String, APIResult>();
         final Set<String> colos = getApplicableColos(type, entity);
 
+        doesEntityHasExtensionJobTag(entity);
```
pallavi-rao: Nit : Method name, entityHasExtensionJobTag, is good enough.

prism/src/main/java/org/apache/falcon/resource/proxy/SchedulableEntityManagerProxy.java
```diff
@@ -310,6 +319,12 @@ public APIResult update(
             @Dimension("colo") @QueryParam("colo") String ignore,
             @QueryParam("skipDryRun") final Boolean skipDryRun) {
 
+        try {
+            isEntityPartOfAnExtension(EntityUtil.getEntity(type, entityName));
```
pallavi-rao: Shouldn't you use the supplied entity definition here, rather than fetching it from from Config Store?

prism/src/main/java/org/apache/falcon/resource/proxy/SchedulableEntityManagerProxy.java
```diff
@@ -378,6 +393,34 @@ protected APIResult doExecute(String colo) throws FalconException {
...
+        checkExtensionJobExist(tags);
+    }
+
+
+    private void doesEntityHasExtensionJobTag(Entity entity) {
+        String tags = entity.getTags();
+        if (StringUtils.isNotBlank(tags)) {
+            String jobName = ExtensionManager.getJobNameFromTag(tags);
+            if (StringUtils.isNotBlank(jobName)) {
+                throw FalconWebException.newAPIException("Entity submit is not allowed on an entity having extension "
```
pallavi-rao: The error message can also say, that such entities need to be submitted as extension jobs.

prism/src/main/java/org/apache/falcon/resource/proxy/SchedulableEntityManagerProxy.java
sandeepSamudrala: The new definition can/might have removed the tag information, hence old entity needs to be used.

prism/src/main/java/org/apache/falcon/resource/proxy/SchedulableEntityManagerProxy.java
sandeepSamudrala: Ack. Will make the change.

prism/src/main/java/org/apache/falcon/resource/proxy/SchedulableEntityManagerProxy.java
sandeepSamudrala: Ack. Will make the change.

prism/src/main/java/org/apache/falcon/resource/proxy/SchedulableEntityManagerProxy.java
sandeepSamudrala: Makes sense, will update the pull req.

PR #318, opened 2016-12-08
FALCON-2211 Falcon Fails to start in distributed mode

PraveenAdlakha:


common/src/main/resources/startup.properties
```diff
@@ -45,6 +44,16 @@
                         org.apache.falcon.service.LogCleanupService,\
                         org.apache.falcon.service.GroupsService,\
                         org.apache.falcon.service.ProxyUserService
+
+
+##Add if you want to use Meatadata Service
+## Also enable all the properties which starts with falcon.graph prefix
+#org.apache.falcon.metadata.MetadataMappingService,\
+
+##Add if you want to use Trusted and NonTrusted Extensions
```
pallavi-rao: Call it Custom or User extensions.  Is it ok if this Service is at the bottom of the list? If not, call out where it needs to appear.

src/conf/startup.properties
```diff
@@ -43,16 +43,22 @@
...
                         org.apache.falcon.service.GroupsService,\
                         org.apache.falcon.service.ProxyUserService
 
+##Add if you want to use Meatadata Service
+## Also enable all the properties which starts with falcon.graph prefix
+#org.apache.falcon.metadata.MetadataMappingService,\
+
+##Add if you want to use Trusted and NonTrusted Extensions
+## In case of distributed Mode enable ExtensionService only on Prism
+#org.apache.falcon.extensions.ExtensionService,\
```
pallavi-rao: Call it Custom or User extensions.  Is it ok if this Service is at the bottom of the list? If not, call out where it needs to appear.

PR #319, opened 2016-12-09
FALCON-2214 Falcon extension registered without fully qualified hdfs …

PraveenAdlakha:
…path

### DIFF
--- a/extensions/src/main/java/org/apache/falcon/extensions/store/ExtensionStore.java
+++ b/extensions/src/main/java/org/apache/falcon/extensions/store/ExtensionStore.java
@@ -269,22 +269,22 @@ public final class ExtensionStore {
         }
     }
 
+    private void assertURI(String part, String value) throws ValidationException {
+        if (value == null) {
+            String msg = "Invalid Path supplied. " + part + " is missing. "
+                    + " Path must contain scheme, authority and path.";
+            LOG.error(msg);
+            throw new ValidationException(msg);
+        }
+    }
+
     public String registerExtension(final String extensionName, final String path, final String description,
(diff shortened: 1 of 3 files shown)


extensions/src/main/java/org/apache/falcon/extensions/store/ExtensionStore.java
```diff
@@ -269,6 +269,18 @@ public String registerExtension(final String extensionName, final String path, f
         throws URISyntaxException, FalconException {
         Configuration conf = new Configuration();
         URI uri = new URI(path);
+        if (uri.getScheme() == null){
```
pallavi-rao: Nit : May be club all 3 checks. And say, "Extension path must contain a scheme, authority and path. Example, hdfs://<namenode>:port/projects/falcon/extension"

PraveenAdlakha: I thought of doing  it..but later though it's better to explicitly tell the error with the path.Don't you think so as we are not saving anything .

pallavi-rao: Actually, yes. It is always better to be as specific as possible. My concern here was more on redundant code. I guess that can be solved by introducing a method like the one below:
```
    private void assertURI(String part, String value) throws ValidationException {
        if (value == null) {
            String msg = "Invalid Path supplied. " + part + " is missing. "
                    + " Path must contain scheme, authority and path.";
            LOG.error(msg);
            throw new ValidationException(msg);
        }
    }
```

PR #320, opened 2016-12-13
FALCON-2219 Untracked files after falcon build

PraveenAdlakha:


.gitignore
```diff
@@ -42,3 +42,6 @@ logs
 falcon-ui/dist
 falcon-ui/node
 falcon-ui/node_modules
+
+#Resources jar
+unit/src/test/resources/falcon-extensions-0.11-SNAPSHOT-tests.jar
```
PracheerAgarwal: can we remove the version number from added entry?

PraveenAdlakha: yes should do that.

PR #325, opened 2016-12-22
FALCON-2014 Fixing Idempotency issue in Distributed mode

sandeepSamudrala:


prism/src/main/java/org/apache/falcon/resource/AbstractEntityManager.java
```diff
@@ -266,28 +265,33 @@ protected APIResult delete(String type, String entity, String colo) {
         try {
             EntityType entityType = EntityType.getEnum(type);
             String removedFromEngine = "";
+            Entity entityObj;
             try {
-                Entity entityObj = EntityUtil.getEntity(type, entity);
-                verifySafemodeOperation(entityObj, EntityUtil.ENTITY_OPERATION.DELETE);
+                entityObj = EntityUtil.getEntity(type, entity);
+            } catch (FalconException e) {
```
pallavi-rao: This is no different from what is happening currently. org.apache.falcon.entity.EntityUtil#getEntity(org.apache.falcon.entity.v0.EntityType, java.lang.String) throws a EntityNotRegisteredException which is a sub-class of FalconException.

ajayyadava: @sandeepSamudrala It will be helpful to review if you could tell the root cause of the issue

sandeepSamudrala: Closing this pull request as its not actually fixing the issue and the my assumption of failure is not right. Will raise a fresh pull request once I have more findings.

PR #326, opened 2016-12-22
FALCON-2227 Fixing error throws to throw exceptions and also added a check for META-INF/services file to be present in the extensions Jar in the extension build path

sandeepSamudrala:


client/src/main/java/org/apache/falcon/client/FalconClient.java
```diff
@@ -84,57 +84,55 @@
...
-    public static final String ORDER_BY = "orderBy";
-    public static final String SORT_ORDER = "sortOrder";
-    public static final String OFFSET = "offset";
-    public static final String NUM_RESULTS = "numResults";
-    public static final String START = "start";
-    public static final String END = "end";
-    public static final String INSTANCE_TIME = "instanceTime";
-    public static final String INSTANCE_STATUS = "instanceStatus";
-    public static final String PROPERTIES = "properties";
+    private static final String RUN_ID = "runid";
```
pallavi-rao: Nit: Why has the access been changed? Ideally, these should be the ones used by FalconCLI too, rather than having these constants repeated FalconCLIConstants.

client/src/main/java/org/apache/falcon/ExtensionHandler.java
```diff
@@ -106,6 +108,10 @@
     public static List<Entity> prepare(String extensionName, String jobName, InputStream configStream, List<URL> urls)
         throws IOException, FalconException {
         ClassLoader extensionClassLoader = ExtensionClassLoader.load(urls);
+        if (extensionClassLoader.getResourceAsStream(EXTENSION_BUILDER_INTERFACE_SERVICE_FILE) == null) {
+            throw new FalconCLIException("The jars in the extension build are not packaged with META-INF/services/"
```
pallavi-rao: Nit : You can say - "The extension build time jars do not contain ...."

client/src/main/java/org/apache/falcon/client/FalconClient.java
sandeepSamudrala: Its being used in a lot of places and the usage context is different. Not making any changes to keep it in 1 place.

pallavi-rao:
+1. Build failure is unrelated. Will merge shortly.

PR #327, opened 2016-12-22
FALCON-2226 Submit ,Schedule and submitAndSchedule API for extension

PraveenAdlakha:


prism/src/main/java/org/apache/falcon/resource/proxy/ExtensionManagerProxy.java
```diff
@@ -317,8 +330,8 @@ public APIResult submit(
 
         try {
             entityMap = getEntityList(extensionName, jobName, feedForms, processForms, config);
-            submitEntities(extensionName, doAsUser, jobName, entityMap, config);
-        } catch (FalconException | IOException e) {
+            submitEntities(extensionName, jobName, entityMap, config, request);
```
pallavi-rao: Why has doAsUser been knocked off? I know we don't use it now. But, since the plan to enable it later, we shouldn't be removing it.

prism/src/main/java/org/apache/falcon/resource/proxy/ExtensionManagerProxy.java
```diff
@@ -380,51 +400,83 @@ public APIResult submitAndSchedule(
...
-            submitInternal(feed, doAsUser);
-            feedNames.add(feed.getName());
-        }
-        for (Entity process: processes) {
-            submitInternal(process, doAsUser);
-            processNames.add(process.getName());
+
+        for(Map.Entry<EntityType, List<Entity>> entry : entityMap.entrySet()){
+            for(final Entity entity : entry.getValue()){
+                final HttpServletRequest httpServletRequest = getEntityStream(entity, entity.getEntityType(), request);
```
pallavi-rao: Since the eventual goal is to get the bufferedRequest I think getEntityStream and getBufferedRequest into a single method.

prism/src/main/java/org/apache/falcon/resource/proxy/ExtensionManagerProxy.java
```diff
@@ -380,51 +400,83 @@ public APIResult submitAndSchedule(
...
         List<String> processNames = new ArrayList<>();
-        for (Entity feed : feeds) {
-            submitInternal(feed, doAsUser);
-            feedNames.add(feed.getName());
-        }
-        for (Entity process: processes) {
-            submitInternal(process, doAsUser);
-            processNames.add(process.getName());
+
+        for(Map.Entry<EntityType, List<Entity>> entry : entityMap.entrySet()){
```
pallavi-rao: The feeds have to be submitted before processes are. Else, the process submission will fail due to non-existence of a feed.

prism/src/main/java/org/apache/falcon/resource/proxy/ExtensionManagerProxy.java
```diff
@@ -691,4 +778,61 @@ private static void checkIfExtensionServiceIsEnabled() {
                     ExtensionService.SERVICE_NAME + " is not enabled.", Response.Status.NOT_FOUND);
         }
     }
+
+    private abstract class EntitiesProxy<T extends APIResult> {
```
pallavi-rao: Why is this class here? Where is it used?

unit/src/main/java/org/apache/falcon/unit/LocalExtensionManager.java
```diff
@@ -18,35 +18,69 @@
...
+public class LocalExtensionManager extends ExtensionManagerProxy {
     public LocalExtensionManager() {}
 
     public APIResult submitExtensionJob(String extensionName, String jobName, InputStream config,
                                         Map<EntityType, List<Entity>> entityMap) throws FalconException, IOException {
-        submitEntities(extensionName, null, jobName, entityMap, config);
+
+        for(Map.Entry<EntityType, List<Entity>> entry : entityMap.entrySet()){
+            for(Entity entity : entry.getValue()){
+                submitInternal(entity, "falconUser");
```
pallavi-rao: Feeds before processes, here too. May be create a method and call from both submit and submitAndSchedule.

unit/src/main/java/org/apache/falcon/unit/LocalExtensionManager.java
```diff
@@ -18,35 +18,69 @@
...
+    public APIResult submitAndSchedulableExtensionJob(String extensionName, String jobName, InputStream configStream,
                                                       Map<EntityType, List<Entity>> entityMap)
         throws FalconException, IOException {
-        submitEntities(extensionName, null, jobName, entityMap, config);
-        scheduleEntities(entityMap);
+        List<String> feedNames = new ArrayList<>();
+        List<String> processNames = new ArrayList<>();
+        for(Map.Entry<EntityType, List<Entity>> entry : entityMap.entrySet()){
+            for(Entity entity : entry.getValue()){
+                submitInternal(entity, "falconUser");
```
pallavi-rao: feeds before processes.

unit/src/main/java/org/apache/falcon/unit/LocalExtensionManager.java
```diff
@@ -18,35 +18,69 @@
...
 import java.io.InputStream;
+import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
 
 /**
  * A proxy implementation of the extension operations in local mode.
  */
-public class LocalExtensionManager extends AbstractExtensionManager {
+public class LocalExtensionManager extends ExtensionManagerProxy {
```
pallavi-rao: Why extend Proxy? Falcon Unit shouldn't worry about Proxy.

prism/src/main/java/org/apache/falcon/resource/proxy/ExtensionManagerProxy.java
PraveenAdlakha: submit API in SchedulableEntityManagerProxy doesnot take doAs as a param we should not do it too.

prism/src/main/java/org/apache/falcon/resource/proxy/ExtensionManagerProxy.java
PraveenAdlakha: not needed will remove.

prism/src/main/java/org/apache/falcon/resource/proxy/ExtensionManagerProxy.java
```diff
@@ -326,7 +325,7 @@ public APIResult submit(
             @FormDataParam("feeds") List<FormDataBodyPart> feedForms,
             @FormDataParam("config") InputStream config) {
         checkIfExtensionServiceIsEnabled();
-        Map<EntityType, List<Entity>> entityMap;
+        TreeMap<EntityType, List<Entity>> entityMap;
```
pallavi-rao: Declaration should only have Map... Only instantiation should be TreeMap.

prism/src/main/java/org/apache/falcon/resource/proxy/ExtensionManagerProxy.java
```diff
@@ -338,15 +337,15 @@ public APIResult submit(
         return new APIResult(APIResult.Status.SUCCEEDED, "Extension job submitted successfully:" + jobName);
     }
 
-    private Map<EntityType, List<Entity>> getEntityList(String extensionName, String jobName,
+    private TreeMap<EntityType, List<Entity>> getEntityList(String extensionName, String jobName,
```
pallavi-rao: Declaration should just be Map

prism/src/main/java/org/apache/falcon/resource/proxy/ExtensionManagerProxy.java
```diff
@@ -338,15 +337,15 @@ public APIResult submit(
...
                                                         List<FormDataBodyPart> feedForms,
                                                         List<FormDataBodyPart> processForms, InputStream config)
         throws FalconException, IOException{
         List<Entity> processes = getProcesses(processForms);
         List<Entity> feeds = getFeeds(feedForms);
         ExtensionType extensionType = getExtensionType(extensionName);
         List<Entity> entities;
-        Map<EntityType, List<Entity>> entityMap = new HashMap<>();
+        TreeMap<EntityType, List<Entity>> entityMap = new TreeMap<>();
```
pallavi-rao: Believe you made this a tree map, so that during submission, feeds go first? But, while putting elements into this map, processes are going in first.

prism/src/main/java/org/apache/falcon/resource/proxy/ExtensionManagerProxy.java
```diff
@@ -442,7 +441,7 @@ private BufferedRequest getBufferedRequest(HttpServletRequest request) {
     }
 
     protected void submitEntities(String extensionName, String jobName,
-                                  Map<EntityType, List<Entity>> entityMap, InputStream configStream,
+                                  TreeMap<EntityType, List<Entity>> entityMap, InputStream configStream,
```
pallavi-rao: Method signature should just have the interface name, Map.

unit/src/main/java/org/apache/falcon/unit/FalconUnitClient.java
```diff
@@ -285,14 +285,14 @@ public APIResult submitExtensionJob(String extensionName, String jobName, String
 
         InputStream configStream = getServletInputStream(configPath);
         try {
-            Map<EntityType, List<Entity>> entityMap = getEntityTypeListMap(extensionName, jobName, configStream);
+            TreeMap<EntityType, List<Entity>> entityMap = getEntityTypeListMap(extensionName, jobName, configStream);
```
pallavi-rao: Same comment as above for the declarations in this class too. Method signatures and declaration should just have Map.

prism/src/main/java/org/apache/falcon/resource/proxy/ExtensionManagerProxy.java
PraveenAdlakha: I want the contract to be tightly coupled so have changed it to SortedMap.

prism/src/main/java/org/apache/falcon/resource/proxy/ExtensionManagerProxy.java
PraveenAdlakha: Internally tree will store in the natural ordering and Enum maintains it in the ordinal value in which feed it the first one so we should be good here.

prism/src/main/java/org/apache/falcon/resource/proxy/ExtensionManagerProxy.java
PraveenAdlakha: changed it to SortedMap.

prism/src/main/java/org/apache/falcon/resource/proxy/ExtensionManagerProxy.java
pallavi-rao: Oh.. right.. My bad. In my head, I mixed it up with LinkedHashMap.

prism/src/main/java/org/apache/falcon/resource/proxy/ExtensionManagerProxy.java
pallavi-rao: fair enough.

prism/src/main/java/org/apache/falcon/resource/proxy/ExtensionManagerProxy.java
pallavi-rao: Fair enough, as long as it is an Interface as you have used (SortedMap) and not the impl., it is fine.

pallavi-rao:
Unrelated build failure.. merging.

PR #328, opened 2016-12-25
FALCON-2197, enable and disable functionalities for extensions

PracheerAgarwal:


client/src/main/java/org/apache/falcon/client/FalconClient.java
```diff
@@ -358,7 +358,9 @@ public void setDebugMode(boolean debugMode) {
         UNREGISTER("api/extension/unregister/", HttpMethod.POST, MediaType.TEXT_PLAIN),
         DETAIL("api/extension/detail/", HttpMethod.GET, MediaType.APPLICATION_JSON),
         JOB_DETAILS("api/extension/extensionJobDetails/", HttpMethod.GET, MediaType.APPLICATION_JSON),
-        REGISTER("api/extension/register/", HttpMethod.POST, MediaType.TEXT_PLAIN);
+        REGISTER("api/extension/register/", HttpMethod.POST, MediaType.TEXT_PLAIN),
+        ENABLE("api/extension/update", HttpMethod.POST, MediaType.TEXT_XML),
```
sandeepSamudrala: change api query path to enable and disable

extensions/src/main/java/org/apache/falcon/extensions/store/ExtensionStore.java
```diff
@@ -346,4 +347,32 @@ public boolean isExtensionStoreInitialized() {
...
+        } else {
+            metaStore.updateExtension(extensionName, PersistenceConstants.DISABLE_EXTENSION);
+            return "Disabled extension:" + extensionName;
+        }
+    }
+
+    public String enableExtension(final String extensionName, String currentUser) throws FalconException {
+
+        if (!metaStore.checkIfExtensionExists(extensionName)) {
+            throw new FalconException("Extension:" + extensionName + " is not registered with Falcon.");
```
sandeepSamudrala: You can move these checks to updateExtension so that its not repeated.

prism/src/main/java/org/apache/falcon/resource/AbstractExtensionManager.java
```diff
@@ -177,4 +177,22 @@ public static String getJobNameFromTag(String tags) {
         }
         return tags.substring(nameStart, nameEnd);
     }
+
+    public String disableExtension(String extensionName, String currentUser) {
+        validateExtensionName(extensionName);
+        try {
+            return ExtensionStore.get().disableExtension(extensionName, currentUser);
```
sandeepSamudrala: You can remove enable and disableExtension and directly call updateExtension with enable or disable as parameter

client/src/main/java/org/apache/falcon/client/FalconClient.java
pallavi-rao: Also, MediaType is TEXT_PLAIN in this case.

common/src/main/java/org/apache/falcon/persistence/ExtensionBean.java
```diff
@@ -44,7 +44,9 @@
         @NamedQuery(name = PersistenceConstants.GET_ALL_EXTENSIONS, query = "select OBJECT(a) from ExtensionBean a "),
         @NamedQuery(name = PersistenceConstants.DELETE_EXTENSIONS_OF_TYPE, query = "delete from ExtensionBean a where a.extensionType = :extensionType "),
         @NamedQuery(name = PersistenceConstants.DELETE_EXTENSION, query = "delete from ExtensionBean a where a.extensionName = :extensionName "),
-        @NamedQuery(name = PersistenceConstants.GET_EXTENSION, query = "select OBJECT(a) from ExtensionBean a where a.extensionName = :extensionName")
+        @NamedQuery(name = PersistenceConstants.GET_EXTENSION, query = "select OBJECT(a) from ExtensionBean a where a.extensionName = :extensionName"),
+        @NamedQuery(name = PersistenceConstants.ENABLE_EXTENSION, query = "update ExtensionBean OBJECT(a) set is_enabled = true where a.extensionName = :extensionName"),
```
pallavi-rao: Having a status in the DB is far more extensible that having a boolean. Else, it will involve DB migration later. You can have a simple enum with the ENABLED and DISABLED as statuses. And, store them in the DB.

extensions/src/main/java/org/apache/falcon/extensions/store/ExtensionStore.java
```diff
@@ -346,4 +347,32 @@ public boolean isExtensionStoreInitialized() {
         return (storePath != null);
     }
 
+    public String disableExtension(final String extensionName, String currentUser) throws FalconException {
+        ExtensionType extensionType = AbstractExtension.isExtensionTrusted(extensionName) ? ExtensionType.TRUSTED
+                : ExtensionType.CUSTOM;
+        if (extensionType.equals(ExtensionType.TRUSTED)) {
+            throw new ValidationException(extensionName + " is trusted cannot be disabled.");
```
pallavi-rao: Nit : Punctuate - extensionName + " is trusted. cannot be disabled."

extensions/src/main/java/org/apache/falcon/extensions/store/ExtensionStore.java
pallavi-rao: I wouldn't recommend moving it to updateExtension. That method is generic enough to take any update.  Instead, have a private validate method that can be invoked by both enable and disable.

extensions/src/main/java/org/apache/falcon/extensions/store/ExtensionStore.java
```diff
@@ -346,4 +347,32 @@ public boolean isExtensionStoreInitialized() {
...
+        } else if (!metaStore.checkIfExtensionExists(extensionName)) {
+            throw new FalconException("Extension:" + extensionName + " is not registered with Falcon.");
+        } else if (!metaStore.getDetail(extensionName).getExtensionOwner().equals(currentUser)) {
+            throw new FalconException("User: " + currentUser + " is not allowed to disable extension: "
+                    + extensionName);
+        } else {
```
pallavi-rao: Do you also want to check if it is already disabled and show an appropriate message?

extensions/src/main/java/org/apache/falcon/extensions/store/ExtensionStore.java
```diff
@@ -346,4 +347,32 @@ public boolean isExtensionStoreInitialized() {
...
+        }
+    }
+
+    public String enableExtension(final String extensionName, String currentUser) throws FalconException {
+
+        if (!metaStore.checkIfExtensionExists(extensionName)) {
+            throw new FalconException("Extension:" + extensionName + " is not registered with Falcon.");
+        } else if (!metaStore.getDetail(extensionName).getExtensionOwner().equals(currentUser)) {
+            throw new FalconException("User: " + currentUser + " is not allowed to enable extension: " + extensionName);
+        } else {
```
pallavi-rao: Do you also want to check if it is already enabled and show an appropriate message?

prism/src/main/java/org/apache/falcon/resource/AbstractExtensionManager.java
pallavi-rao: Hmm... updateExtensionStatus, may be...

client/src/main/java/org/apache/falcon/client/FalconClient.java
```diff
@@ -358,7 +358,9 @@ public void setDebugMode(boolean debugMode) {
         UNREGISTER("api/extension/unregister/", HttpMethod.POST, MediaType.TEXT_PLAIN),
         DETAIL("api/extension/detail/", HttpMethod.GET, MediaType.APPLICATION_JSON),
         JOB_DETAILS("api/extension/extensionJobDetails/", HttpMethod.GET, MediaType.APPLICATION_JSON),
-        REGISTER("api/extension/register/", HttpMethod.POST, MediaType.TEXT_PLAIN);
+        REGISTER("api/extension/register/", HttpMethod.POST, MediaType.TEXT_PLAIN),
+        ENABLE("api/extension/enable", HttpMethod.POST, MediaType.TEXT_PLAIN),
```
pallavi-rao: Guess this will also be TEXT_XML since we are returning APIResult. Sorry, I misled you earlier.

pallavi-rao:
@PracheerAgarwal, can you please squash your last four commits into 1. It is easy to a diff between my last review the the changes. Else, I'll have to look at the whole patch again OR look at each commit individually.

extensions/src/main/java/org/apache/falcon/extensions/jdbc/ExtensionMetaStore.java
```diff
@@ -154,7 +157,7 @@ public void storeExtensionJob(String jobName, String extensionName, List<String>
         EntityManager entityManager = getEntityManager();
         try {
             beginTransaction(entityManager);
-            entityManager.persist(extensionJobsBean);
+            entityManager.merge(extensionJobsBean);
```
pallavi-rao: Why has been changed to merge?As I understand, merge allows for updates. In this particular API, we only want to allow creation. Isn't it?

PracheerAgarwal: https://github.com/apache/falcon/pull/328/files#diff-b03505aec96486daeb519e68aba63d2dR445

While submitting an extension job, its process and feeds are submit in a sequential manner, so I have made this merge to support update of process and feed list.

pallavi-rao: But, the store is updated only once, irrespective of how many processes and feeds are are in job, isn't it?

PracheerAgarwal: We have changed it to update after every entity submission.

pallavi-rao: ok. Had missed that change.

extensions/src/main/java/org/apache/falcon/extensions/store/ExtensionStore.java
```diff
@@ -346,4 +347,30 @@ public boolean isExtensionStoreInitialized() {
         return (storePath != null);
     }
 
+    public String updateExtensionStatus(final String extensionName, String currentUser, ExtensionStatus status) throws
+            FalconException {
+        validateStatusChange(extensionName, currentUser);
+        if (metaStore.getDetail(extensionName).getStatus().equals(status)) {
+            throw new ValidationException(extensionName + " is already in " + status.toString() + " state.");
```
sandeepSamudrala: why throw error? just send out the message saying its already in the required state.

extensions/src/main/java/org/apache/falcon/extensions/store/ExtensionStore.java
```diff
@@ -346,4 +347,30 @@ public boolean isExtensionStoreInitialized() {
...
+    private void validateStatusChange(final String extensionName, String currentUser) throws FalconException {
+
+        ExtensionType extensionType = AbstractExtension.isExtensionTrusted(extensionName) ? ExtensionType.TRUSTED
+                : ExtensionType.CUSTOM;
+        if (extensionType.equals(ExtensionType.TRUSTED)) {
+            throw new ValidationException(extensionName + " is trusted. Status can't be changed for trusted "
+                    + "extensions.");
+        } else if (!metaStore.checkIfExtensionExists(extensionName)) {
+            throw new FalconException("Extension:" + extensionName + " is not registered with Falcon.");
+        } else if (!metaStore.getDetail(extensionName).getExtensionOwner().equals(currentUser)) {
```
sandeepSamudrala: rather than checking at a user specific? can you update to check if they both belong to same group?
If its making too complex we can live with this.

prism/src/main/java/org/apache/falcon/resource/proxy/ExtensionManagerProxy.java
```diff
@@ -195,6 +197,7 @@ public ExtensionInstanceList getInstances(
     public APIResult schedule(@PathParam("job-name") String jobName,
                               @DefaultValue("") @QueryParam("doAs") String doAsUser) {
         checkIfExtensionServiceIsEnabled();
+        checkIfExtensionIsEnabled(ExtensionStore.getMetaStore().getExtensionJobDetails(jobName).getExtensionName());
```
sandeepSamudrala: should schedule need it as the extension is submitted and entities are submitted?

sandeepSamudrala: Also checkIfExtensionJobExists may be required in schedule too, if wrong jobName is being supplied.

prism/src/main/java/org/apache/falcon/resource/proxy/ExtensionManagerProxy.java
```diff
@@ -460,15 +473,9 @@ protected APIResult doExecute(String colo) throws FalconException {
                 }else{
                     processNames.add(entity.getName());
                 }
+                metaStore.storeExtensionJob(jobName, extensionName, feedNames, processNames, configBytes);
```
sandeepSamudrala: why store is required in two places. its already stored before submitting entities

prism/src/main/java/org/apache/falcon/resource/proxy/ExtensionManagerProxy.java
```diff
@@ -460,15 +473,9 @@ protected APIResult doExecute(String colo) throws FalconException {
                 }else{
```
sandeepSamudrala: Move this part of getting entity Names to top.

sandeepSamudrala: in case there are partial updates the entities information is not stored and can't be deleted.

extensions/src/main/java/org/apache/falcon/extensions/store/ExtensionStore.java
PracheerAgarwal: it was discussed for now, we will check with just user.

prism/src/main/java/org/apache/falcon/resource/proxy/ExtensionManagerProxy.java
PracheerAgarwal: yes, i think schedule should also have that check. Also scheduled extension jobs will be exempted.

extensions/src/main/java/org/apache/falcon/extensions/store/ExtensionStore.java
PracheerAgarwal: I recollect from the discussion we had, we decided asked to throw an error. I also stated the example of current status change of process/feed, where we do not throw an error.

prism/src/main/java/org/apache/falcon/resource/proxy/ExtensionManagerProxy.java
PracheerAgarwal: In the first place, we have just entered empty list of process and feeds.  This is the step where we update that information after every successful entity submission.

sandeepSamudrala:
One comment which has been missed , which I had mentioned earlier too.
Can you make changes to localExtensionManager and modify org.apache.falcon.unit.TestFalconUnit#testSubmitAndScheduleExtensionJob 
to test for disabling and then update and enable it back and to check if it gets updated then

PR #331, opened 2016-12-28
FALCON-2200 Update API support for extension job (user extension)

sandeepSamudrala:


prism/src/main/java/org/apache/falcon/resource/proxy/ExtensionManagerProxy.java
```diff
@@ -472,6 +478,92 @@ protected APIResult doExecute(String colo) throws FalconException {
...
+        List<Entity> feeds = entityMap.get(EntityType.FEED);
+        List<Entity> processes = entityMap.get(EntityType.PROCESS);
+        validateFeeds(feeds);
+        validateProcesses(processes);
+        List<String> feedNames = new ArrayList<>();
+        List<String> processNames = new ArrayList<>();
+
+        for (Map.Entry<EntityType, List<Entity>> entry : entityMap.entrySet()) {
+            for (final Entity entity : entry.getValue()) {
+                final String entityType = entity.getEntityType().toString();
```
pallavi-rao: This repeats code from SchedulableEntityManagerProxy. Any changes/fixes will need to be made to 2 places. Can you consider introducing a Util class and using it in both SchedulableEntityManagerProxy and ExtensionManagerProxy? The problem existed when submit and submitAndSchedule were introduced. As we introduce more APIs, have repetition will get worse.

sandeepSamudrala: made the chcanges accordingly

prism/src/main/java/org/apache/falcon/resource/proxy/ExtensionManagerProxy.java
```diff
@@ -96,8 +96,9 @@
 
     private boolean embeddedMode = DeploymentUtil.isEmbeddedMode();
     private String currentColo = DeploymentUtil.getCurrentColo();
-    private final Map<String, Channel> configSyncChannels = new HashMap<String, Channel>();
-    private final Map<String, Channel> entityManagerChannels = new HashMap<String, Channel>();
+    private final Map<String, Channel> configSyncChannels = new HashMap<>();
```
pallavi-rao: Why do we still need these 2 variables if we are using the the EntityProxyUtil?

sandeepSamudrala: These are not required. Missed to removed them.

sandeepSamudrala:
I have tested them in distributed mode. For embedded mode the unit tests run from FalconCLIIT so that should be sufficient

PR #333, opened 2016-12-28
FALCON-2199 Delete API support for extension job (user extension)

sandeepSamudrala:
This pull request is dependent on https://github.com/apache/falcon/pull/331


prism/src/main/java/org/apache/falcon/resource/proxy/ExtensionManagerProxy.java
```diff
@@ -273,29 +275,67 @@ public APIResult resume(@PathParam("job-name") String jobName,
...
+        return new APIResult(APIResult.Status.SUCCEEDED, "Extension job " + jobName + " deleted successfully");
+    }
+
+    private SortedMap<EntityType,List<Entity>> getJobEntities(String jobName, ExtensionJobsBean extensionJobsBean,
+                                                              String doAsUser) throws FalconException, IOException {
+        TreeMap<EntityType, List<Entity>> entityMap = new TreeMap<>();
+        ExtensionMetaStore metaStore = ExtensionStore.getMetaStore();
+
+        ExtensionType extensionType =  metaStore.getDetail(extensionJobsBean.getExtensionName()).getExtensionType();
+        if (ExtensionType.TRUSTED.equals(extensionType)) {
```
pallavi-rao: Why do this? The list of processes and feeds are stored in the Jobs table, even for trusted extensions.

prism/src/main/java/org/apache/falcon/resource/proxy/ExtensionManagerProxy.java
```diff
@@ -427,6 +467,28 @@ private BufferedRequest getBufferedRequest(HttpServletRequest request) {
         return new BufferedRequest(request);
     }
 
+
+    private void deleteEntities(SortedMap<EntityType, List<Entity>> entityMap, HttpServletRequest request)
+        throws IOException, JAXBException {
+        for (Map.Entry<EntityType, List<Entity>> entry : entityMap.entrySet()) {
+            for (final Entity entity : entry.getValue()) {
+                final HttpServletRequest bufferedRequest = getEntityStream(entity, entity.getEntityType(), request);
```
pallavi-rao: Move this also to the Util method as suggested in PR 331.

sandeepSamudrala: Ack. Made the changes.

sandeepSamudrala:
@pallavi-rao : Apologies . There were more changes into this which have been found during dev testing. I have made more changes. Please re-review it again. I have incorporated your review comments.

client/src/main/java/org/apache/falcon/ExtensionHandler.java
```diff
@@ -185,8 +185,6 @@ private static String createStagePath(String extensionName, String jobName) {
                 for (File innerFile : files) {
                     if (innerFile.isFile()) {
                         urls.add(innerFile.toURI().toURL());
```
pallavi-rao: Don't want to support sub-directories under the build lib path? It is fine to take that stance, but, we will have to call that out explicitly in the documentation.

prism/src/main/java/org/apache/falcon/service/BacklogMetricEmitterService.java
```diff
@@ -116,11 +116,14 @@ public void onRemove(Entity entity) throws FalconException{
         if (entity.getEntityType() != EntityType.PROCESS){
```
pallavi-rao: The check here could cause issues. Entity can be cluster or datasource. In which case, the explicit cast below will cause issues.

unit/src/test/java/org/apache/falcon/unit/TestFalconUnit.java
```diff
@@ -436,34 +438,37 @@ public void testSubmitAndScheduleExtensionJob() throws Exception {
...
-        process = (Process)getClient().getDefinition(EntityType.PROCESS.toString(), "sample", null);
+        process = (Process) getClient().getDefinition(EntityType.PROCESS.toString(), "sample", null);
         Assert.assertEquals(process.getPipelines(), "testSample");
 
         apiResult = getClient().getStatus(EntityType.PROCESS, "sample", CLUSTER_NAME, null, false);
         assertStatus(apiResult);
         Assert.assertEquals(apiResult.getMessage(), "RUNNING");
+
+        apiResult = deleteExtensionJob(TEST_JOB, null);
+        assertStatus(apiResult);
```
pallavi-rao: Also, add asserts to ensure the Job nor its corresponding entities exist.

client/src/main/java/org/apache/falcon/ExtensionHandler.java
sandeepSamudrala: Yes. There are occasional build failures too for recursive calling. I will add it to the documentation.

prism/src/main/java/org/apache/falcon/service/BacklogMetricEmitterService.java
sandeepSamudrala: Backlog is supported only for processes and so this check is there.

unit/src/test/java/org/apache/falcon/unit/TestFalconUnit.java
sandeepSamudrala: Added the cases for the same.

sandeepSamudrala:
Added Assert.fail post the api calls.

PR #335, opened 2017-01-03
FALCON-2229 BacklogEmitterMetricService fix for deleting entities' instances in case of removal of SLA or deletion of entity

sandeepSamudrala:


prism/src/main/java/org/apache/falcon/jdbc/BacklogMetricStore.java
```diff
@@ -70,18 +70,19 @@ public synchronized void deleteMetricInstance(String entityName, String cluster,
...
         }
     }
 
-    public void deleteEntityInstance(String entityName){
+    public void deleteEntityInstance(String entityName, String entityType) {
         EntityManager entityManager = getEntityManager();
         beginTransaction(entityManager);
         Query q = entityManager.createNamedQuery(PersistenceConstants.DELETE_ALL_BACKLOG_ENTITY_INSTANCES);
         q.setParameter("entityName", entityName);
+        q.setParameter("entityType", entityType);
```
PracheerAgarwal: do we need to add clusterName also here?

prism/src/main/java/org/apache/falcon/jdbc/BacklogMetricStore.java
```diff
@@ -70,18 +70,19 @@ public synchronized void deleteMetricInstance(String entityName, String cluster,
...
-        try{
+        try {
             q.executeUpdate();
         } finally {
             commitAndCloseTransaction(entityManager);
         }
     }
 
-    public void deleteEntityInstance(String entityName){
+    public void deleteEntityInstance(String entityName, String entityType) {
```
PracheerAgarwal: can we change the method name to deleteEntityBackLogInstances?

prism/src/main/java/org/apache/falcon/jdbc/BacklogMetricStore.java
sandeepSamudrala: no. In a colo on removal of an entity, its removed from all the clusters, which means deletion of the instances across the clusters per colo.

PracheerAgarwal: but the query for the PersistenceConstants  DELETE_ALL_BACKLOG_ENTITY_INSTANCES expects three variables.
query = delete from BacklogMetricBean a where a.entityName = :entityName and a.clusterName = :clusterName and a.entityType = :entityType"
Which means from there you need to remove clusterName.

sandeepSamudrala: Thanks for catching it. I had missed it. I will remove cluster from the named query.

PracheerAgarwal:
+1

PR #340, opened 2017-01-10
FALCON-2242 Logging Extension Builder's exceptions to client logs

sandeepSamudrala:


client/src/main/java/org/apache/falcon/client/FalconClient.java
```diff
@@ -1155,7 +1163,8 @@ private JSONObject getExtensionJobDetailJson(String jobName) {
                 entities = ExtensionHandler.loadAndPrepare(extensionName, jobName, configStream,
                         extensionBuildLocation);
             } catch (Exception e) {
-                throw new FalconCLIException("Error in building the extension", e);
+                LOG.error("Error in building the extension. Cause: ", e);
```
pallavi-rao: May be you should also log this in OUT stream too as the error might come from the extension builder and the error stack will be easily available to the user.

sandeepSamudrala: Added out to the cli user.

client/src/main/java/org/apache/falcon/client/FalconClient.java
```diff
@@ -1155,7 +1164,9 @@ private JSONObject getExtensionJobDetailJson(String jobName) {
...
+                OUT.get().println("Error in building the extension:" + Arrays.toString(e.getStackTrace()));
```
pallavi-rao: You can use ExceptionUtils.getFullStackTrace (of commons-lang) instead.

PR #343, opened 2017-01-18
FALCON-2260 Enhance extension List api to support for user extensions

sandeepSamudrala:


webapp/src/test/java/org/apache/falcon/resource/ExtensionManagerIT.java
```diff
@@ -100,8 +100,6 @@ public void testTrustedExtensionJob() throws Exception {
         // validate job list results
         ExtensionJobList jobs = context.getExtensionJobs("hdfs-mirroring", null, null, null, null, null);
         Assert.assertEquals(jobs.numJobs, 2);
-        Assert.assertEquals(jobs.job.get(0).jobName, JOB_NAME_1);
```
PraveenAdlakha: Can we unable this test case and see if test cases are running correctly.Right now the test cases have been disabled.

@Test (enabled = false)

extensions/src/main/java/org/apache/falcon/extensions/jdbc/ExtensionMetaStore.java
```diff
@@ -132,6 +132,18 @@ public ExtensionBean getDetail(String extensionName) {
         }
     }
 
+    public List<ExtensionJobsBean> getJobsForAnExtension(String extensionName) {
```
PraveenAdlakha: Please add unit test for this.

common/src/main/java/org/apache/falcon/persistence/PersistenceConstants.java
```diff
@@ -80,5 +80,6 @@ private PersistenceConstants(){
     public static final String GET_ALL_EXTENSION_JOBS = "GET_ALL_EXTENSION_JOBS";
     public static final String DELETE_EXTENSION_JOB = "DELETE_EXTENSION_JOB";
     public static final String GET_EXTENSION_JOB = "GET_EXTENSION_JOB";
+    public static final String GET_JOBS_FOR_AN_EXTENSION = "GET_JOBS_FOR_AN_EXTENSION";
```
sandeepSamudrala: Was thinking the same initially, but then to get all details I can always create a new query. I have made the changes accordingly.

extensions/src/main/java/org/apache/falcon/extensions/jdbc/ExtensionMetaStore.java
sandeepSamudrala: Ack. Added.

webapp/src/test/java/org/apache/falcon/resource/ExtensionManagerIT.java
sandeepSamudrala: This Test still doesn't seem to be working.

pallavi-rao:
I have no further comments. Regarding, ExtensionManagerIT, I think it would be a good idea to park a JIRA to ensure trusted extension is not regressed.

pallavi-rao:
Just one comment, the description of the list option needs to be changed:
Currently it is: `-list                   List extension jobs and associated entities`

sandeepSamudrala:
@pallavi-rao : I had modified the documentation accordingly and I created the following jira to address the test case.
https://issues.apache.org/jira/browse/FALCON-2262

pallavi-rao:
Thanks @sandeepSamudrala. @PraveenAdlakha, please re-review.

PraveenAdlakha:
+1

PR #344, opened 2017-01-22
FALCON-2246 Api to list all the jobs in the system

PracheerAgarwal:


sandeepSamudrala:
Just add a unit test in FalconUnit. Rest looks fine.

cli/src/main/java/org/apache/falcon/cli/FalconExtensionCLI.java
```diff
@@ -51,6 +51,7 @@
 
     // Extension commands
     private static final String ENUMERATE_OPT = "enumerate";
+    private static final String JOB_LIST_OPT = "jobList";
```
PracheerAgarwal: correctly said so, have revert my changes and have modified the existing list API.

prism/src/main/java/org/apache/falcon/resource/proxy/ExtensionManagerProxy.java
```diff
@@ -106,15 +108,35 @@ public ExtensionJobList getExtensionJobs(
...
+            if (extensionName != null) {
+                extensionJobs = ExtensionStore.getMetaStore().getJobsForAnExtension(extensionName);
+            } else {
+                extensionJobs = ExtensionStore.getMetaStore().getAllExtensionJobs();
+            }
+
             switch (sortOrder.toLowerCase()) {
             case DESCENDING_SORT_ORDER:
-                Collections.sort(jobNames, Collections.reverseOrder(String.CASE_INSENSITIVE_ORDER));
+                Collections.sort(extensionJobs, new Comparator<ExtensionJobsBean>() {
```
pallavi-rao: Instead, can we have ExtensionJobsBean implement Comparable interface?

prism/src/main/java/org/apache/falcon/resource/AbstractExtensionManager.java
```diff
@@ -100,6 +107,39 @@ public APIResult getExtensions() {
...
+            @Override
+            public int compare(ExtensionJobsBean o1, ExtensionJobsBean o2) {
+                return o1.getJobName().compareToIgnoreCase(o2.getJobName());
+            }
+        };
+
+        Map<String, String> jobAndExtensionNames = new HashMap<>();
+        List<ExtensionJobsBean> extensionJobs = null;
+        if (extensionName != null) {
+            extensionJobs = new ArrayList<>(ExtensionStore.getMetaStore().getJobsForAnExtension(extensionName));
```
sandeepSamudrala: what if the result set is 0 or null?

PR #345, opened 2017-01-23
FALCON-2247 include status of entities in getDetailExtensionJobs

PracheerAgarwal:


prism/src/main/java/org/apache/falcon/resource/AbstractExtensionManager.java
```diff
@@ -142,8 +146,29 @@ private JSONObject buildExtensionJobDetailResult(final String jobName) throws Fa
...
             detailsObject.put(JOB_NAME, jobsBean.getJobName());
             detailsObject.put(EXTENSION_NAME, jobsBean.getExtensionName());
-            detailsObject.put(FEEDS, StringUtils.join(jobsBean.getFeeds(), ","));
-            detailsObject.put(PROCESSES, StringUtils.join(jobsBean.getProcesses(), ","));
+
+            JSONObject feedsObject = new JSONObject();
+            for (String feed : jobsBean.getFeeds()) {
+                try {
+                    EntityUtil.getEntity(EntityType.FEED, feed);
+                } catch (EntityNotRegisteredException e) {
```
sandeepSamudrala: only type is changed in these 2 segments. Extract a method for the same

prism/src/main/java/org/apache/falcon/resource/AbstractExtensionManager.java
```diff
@@ -266,4 +270,17 @@ protected static void checkIfExtensionJobNameExists(String jobName, String exten
...
         }
     }
+
+    private JSONObject getEntitiesStatus(List<String> entities, EntityType type) throws JSONException, FalconException {
+        JSONObject entityObject = new JSONObject();
+        for (String entity : entities) {
+            try {
+                EntityUtil.getEntity(type, entity);
+            } catch (EntityNotRegisteredException e) {
+                entityObject.put(entity, ENTITY_NOT_EXISTS_STATUS);
```
pallavi-rao: By JSON standards, the keys are usually constant strings. This should be `{ "name" : "abc, "status" : "EXISTS"}`

This makes it easy to consume programmatically also.

PracheerAgarwal: valid point. Have made the changes.

PR #348, opened 2017-01-30
FALCON-2220 adding doAsUser parameter in extension APIs

PracheerAgarwal:


PracheerAgarwal:
This is just the skeleton of this PR. Will add one more commit.

client/src/main/java/org/apache/falcon/client/FalconClient.java
```diff
@@ -1016,9 +1016,10 @@ public String getEdge(String id, String doAsUser) {
         return sendMetadataLineageRequest(MetadataOperations.EDGES, id, doAsUser);
     }
 
-    public APIResult enumerateExtensions()  {
+    public APIResult enumerateExtensions(final String doAsUser)  {
```
pallavi-rao: Nothing related to your patch, but, can you add an @Override annotation for all the methods that are implementation of the Abstract class. We have been missing that.

PR #357, opened 2017-02-02
Falcon 2273

vrangan:
Added to code to disable external entity resolution and not log messages with info about location of password files


common-types/src/main/java/org/apache/falcon/entity/v0/Entity.java
```diff
@@ -82,7 +84,11 @@ public String toString() {
     public static Entity fromString(EntityType type, String str) {
         try {
             Unmarshaller unmarshaler = type.getUnmarshaller();
-            return (Entity) unmarshaler.unmarshal(new StringReader(str));
+            XMLInputFactory xif = XMLInputFactory.newFactory();
+            xif.setProperty(XMLInputFactory.IS_SUPPORTING_EXTERNAL_ENTITIES, false);
```
sandeepSamudrala: Can you extract this editing before giving back to user to a method and so can be reused all places.

pallavi-rao:
@vrangan, can you please address or respond to @sandeepSamudrala's comment. I would like this fix to be part of our upcoming 0.11 release.

vrangan:
@pallavi-rao and @sandeepSamudrala 
Updated with review comments.  Please check

pallavi-rao:
Unrelated build failure, merging.

PR #358, opened 2017-02-07
FALCON-2274 Job list in extension

PracheerAgarwal:


PracheerAgarwal:
dev-tested
falcon extension -list
Hadoop is installed, adding hadoop classpath to falcon classpath
SLF4J: Class path contains multiple SLF4J bindings.
SLF4J: Found binding in [jar:file:/usr/local/falcon/falcon-distributed-0.11-SNAPSHOT/client/lib/slf4j-log4j12-1.7.5.jar!/org/slf4j/impl/StaticLoggerBinder.class]
SLF4J: Found binding in [jar:file:/usr/hdp/2.2.4.2-2/hadoop/lib/slf4j-log4j12-1.7.5.jar!/org/slf4j/impl/StaticLoggerBinder.class]
SLF4J: See http://www.slf4j.org/codes.html#multiple_bindings for an explanation.
SLF4J: Actual binding is of type [org.slf4j.impl.Log4jLoggerFactory]
4
sampleAsc    extension-example
sampleAsc2    extension-example

prism/src/main/java/org/apache/falcon/resource/proxy/ExtensionManagerProxy.java
```diff
@@ -94,14 +94,16 @@
 
     //SUSPEND CHECKSTYLE CHECK ParameterNumberCheck
     @GET
-    @Path("list/{extension-name}")
+    @Path("list")
     @Produces({MediaType.TEXT_XML, MediaType.APPLICATION_JSON})
     public ExtensionJobList getExtensionJobs(
-            @PathParam("extension-name") String extensionName,
+            @QueryParam("extension-name") String extensionName,
```
pallavi-rao: Why query param? Path param is REST compliant and is in-line with other APIs too.

PracheerAgarwal: this API is used for two use case, one to get the jobs for an extension and other to get all jobs irrespective of registered extension. 
In the second case, we were not passing anything in extension-name, means null.. which resulted in error from FalconClient that the given parameter is null. So I changed it to QueryParam.

pallavi-rao: Understand the intent. Can you see if we can do something like what we do for entity listing. Entity type is part of the path, but, is optional.

prism/src/main/java/org/apache/falcon/resource/proxy/ExtensionManagerProxy.java
```diff
@@ -94,14 +94,16 @@
...
-    @Path("list/{extension-name}")
+    @Path("list{extension-name : (/[^/]+)?}")
     @Produces({MediaType.TEXT_XML, MediaType.APPLICATION_JSON})
     public ExtensionJobList getExtensionJobs(
             @PathParam("extension-name") String extensionName,
             @DefaultValue(ASCENDING_SORT_ORDER) @QueryParam("sortOrder") String sortOrder,
             @DefaultValue("") @QueryParam("doAs") String doAsUser) {
         checkIfExtensionServiceIsEnabled();
-        getExtensionIfExists(extensionName);
+        if (extensionName != null) {
```
pallavi-rao: Use StringUtils.isNotBlank instead.

prism/src/main/java/org/apache/falcon/resource/proxy/ExtensionManagerProxy.java
```diff
@@ -94,14 +96,16 @@
...
             @PathParam("extension-name") String extensionName,
             @DefaultValue(ASCENDING_SORT_ORDER) @QueryParam("sortOrder") String sortOrder,
             @DefaultValue("") @QueryParam("doAs") String doAsUser) {
         checkIfExtensionServiceIsEnabled();
-        getExtensionIfExists(extensionName);
+        if (StringUtils.isNotBlank(extensionName)) {
+            getExtensionIfExists(extensionName);
+        }
         try {
             return super.getExtensionJobs(extensionName, sortOrder, doAsUser);
```
pallavi-rao: Do you also want to modify getExtensionJobs to use StringUtils.isNotBlank? Because now, extensionName will not be null.

PracheerAgarwal: smart catch. Ack.

PR #360, opened 2017-02-08
FALCON-2276 Falcon Trusted extensions submission and validation was failing

sandeepSamudrala:


client/src/main/java/org/apache/falcon/client/FalconClient.java
```diff
@@ -1169,7 +1169,7 @@ private JSONObject getExtensionJobDetailJson(String jobName) {
     private List<Entity> getEntities(String extensionName, String jobName, InputStream configStream,
                                      String extensionType, String extensionBuildLocation) {
         List<Entity> entities = null;
-        if (!extensionType.equals(ExtensionType.CUSTOM.name())) {
+        if (!extensionType.equals(ExtensionType.TRUSTED.toString())) {
```
pallavi-rao: Shouldn't the comparison be with ExtensionType.TRUSTED.name()? Doesn't ExtensionType.TRUSTED.toString() result in "Trusted extension", rather than "TRUSTED"?

client/src/main/java/org/apache/falcon/client/FalconClient.java
```diff
@@ -1209,7 +1209,7 @@ public APIResult validateExtensionJob(final String extensionName, final String j
                                           final String configPath, final String doAsUser) {
         String extensionType = ExtensionHandler.getExtensionType(extensionName, getExtensionDetailJson(extensionName));
         InputStream configStream = getServletInputStream(configPath);
-        if (ExtensionType.TRUSTED.name().equalsIgnoreCase(extensionType)) {
+        if (extensionType.equals(ExtensionType.TRUSTED.toString())) {
```
pallavi-rao: Same comment as above name() vs. toString().

client/src/main/java/org/apache/falcon/client/FalconClient.java
sandeepSamudrala: This comes as Trusted extension from the server as its .toString while coming from the server.

pallavi-rao: ah.. got it. Thanks for clarifying.

PR #366, opened 2017-02-10
FALCON-2281 HiveDR fails with permissions denied

sowmyaramesh:


peeyushb:
+1. Patch looks good. 
If no further comments, I will commit this patch.

PR #367, opened 2017-02-13
FALCON-2283 Issue with jersey multi part with mime type 1.6 version

sandeepSamudrala:


ajayyadava:
@sandeepSamudrala Can you elaborate on your analysis of what was the underlying cause and how this change fixes the issue?

ajayyadava:
Nevermind. I think I found the probable issue. I believe it is something on these lines. For the benefit of others it seems the issues is: https://java.net/jira/browse/JAX_WS-965 and fix is on the lines of http://www.javaroots.com/2013/09/jersey-rest-service-stream-already.html

Please, try to add as much details as possible for bug fixes. It not only helps in reviewing, but will also help users who might be affected by the same issue. 

Can you please add a test for the same? A test which will fail without the fix and pass with the fix.

sandeepSamudrala:
@ajayyadava : There are test cases for the same in ExtensionManagerIT, but unfortunately they are disabled because of which this was not caught in tests. 
There is a jira already parked for the same 
https://issues.apache.org/jira/browse/FALCON-2262

sandeepSamudrala:
@ajayyadava : Do you still have any concerns? If not would like to get this merged, as this is blocker for the release.

ajayyadava:
+1 Thanks @sandeepSamudrala

pallavi-rao:
Thanks @sandeepSamudrala , @ajayyadava

PR #373, opened 2017-02-23
FALCON-2288, overwriting jobName while job submission for trusted extensions

PracheerAgarwal:


prism/src/main/java/org/apache/falcon/resource/proxy/ExtensionManagerProxy.java
```diff
@@ -308,6 +310,9 @@ public APIResult submit(
                     trustedProcesses.add(entity);
                 }
             }
+            // add tags on extension name and job
+            EntityUtil.applyTags(extensionName, jobName, trustedFeeds);
```
pallavi-rao: Applying the Tags and returning the entityMap can be moved outside if-else to avoid code repetition. Also, not related with patch, but,  getEntityList method itself can be re-organized a little to avoid code repetition.

PR #375, opened 2017-03-03
FALCON-2292 . User Extension submission fails for late input section validation

sandeepSamudrala:


sandeepSamudrala:
LateInputSection validation has to be avoided as the input feeds may not have been submitted by then.

pallavi-rao:
@sandeepSamudrala wouldn't the feeds be submitted before processes? Even if the input feeds are not part of an extension job, it should be submitted before the extension job is submitted, isn't it? What am I missing?

sandeepSamudrala:
@pallavi-rao : Validation for processes/feeds happens prior to submission(by passing input and output feeds checks) and later when the individual entities are submitted, then validation happens for input and output sections of the process

pallavi-rao:
Thanks @sandeepSamudrala for clarifying. 👍

PR #377, opened 2017-03-09
FALCON-2285 Ignored instances are not being abled to be killed

sandeepSamudrala:


oozie/src/main/java/org/apache/falcon/workflow/engine/OozieWorkflowEngine.java
```diff
@@ -949,7 +949,8 @@ private void performAction(String cluster, JobAction action, CoordinatorAction c
                 status = Status.KILLED.name();
                 break;
             }
-            if (!WF_KILL_PRECOND.contains(jobInfo.getStatus())) {
+            if (!WF_KILL_PRECOND.contains(jobInfo.getStatus()) &&
+                    jobInfo.getStatus().name().equalsIgnoreCase(Status.IGNORED.name())) {
```
pallavi-rao: Shouldn't you rather add IGNORED to WF_KILL_PRECOND, instead?

sandeepSamudrala:
Closing this pull request as this is more an issue from oozie as the instance status not being changed to READY from KILLED state upon rerun until as there is a concurrency limit. This has to be fixed at the oozie end.

PR #378, opened 2017-03-22
FALCON-2293 falcon extension fails for deletion if 2 extensions have same entity name 

sandeepSamudrala:


prism/src/main/java/org/apache/falcon/resource/proxy/ExtensionManagerProxy.java
```diff
@@ -503,20 +504,34 @@ public int read() throws IOException {
...
         }
     }
 
+    private void checkIfPartOfAnotherExtension(Entity process, EntityType entityType, String jobName)
+        throws FalconException {
+        try {
+            Entity entity = EntityUtil.getEntity(entityType, process.getName());
+            String extractedJobName = AbstractExtensionManager.getJobNameFromTag(entity.getTags());
+            if (!extractedJobName.equals(jobName)) {
+                throw new FalconException("Entity is part another extension job: " + extractedJobName);
```
pallavi-rao: Message could be slightly misleading. If extractedJobName is null (when the entity is not part of a Job), this will still say it is part of an extension.

prism/src/main/java/org/apache/falcon/resource/proxy/ExtensionManagerProxy.java
```diff
@@ -503,20 +504,34 @@ public int read() throws IOException {
...
-    private void validateProcesses(List<Entity> processes) throws FalconException {
+    private void validateProcesses(List<Entity> processes, String jobName) throws FalconException {
         ProcessEntityParser processEntityParser = new ProcessEntityParser();
         for (Entity process : processes) {
+            checkIfPartOfAnotherExtension(process, EntityType.PROCESS, jobName);
             processEntityParser.validate((Process) process, false);
         }
     }
 
+    private void checkIfPartOfAnotherExtension(Entity process, EntityType entityType, String jobName)
```
pallavi-rao: Nit : variable name should be entity

sandeepSamudrala:
Dev tested :

[sandeep@/usr/local/falcon/falcon-distributed-0.11-SNAPSHOT$]falcon extension -submitAndSchedule -jobName sample -file /tmp/1234  -extensionName extension-example
Extension job submitted and scheduled successfully

[sandeep@/usr/local/falcon/falcon-distributed-0.11-SNAPSHOT$]falcon extension -submitAndSchedule -ERROR: Internal Server Error;Entity sample is part another extension job: sample

unit/src/test/java/org/apache/falcon/unit/TestFalconUnit.java
```diff
@@ -493,6 +493,7 @@ public void testExtensionJobOperations() throws Exception {
         }
```
pallavi-rao: A negative test case to ensure an error is thrown?

unit/src/test/java/org/apache/falcon/unit/TestFalconUnit.java
```diff
@@ -493,6 +493,7 @@ public void testExtensionJobOperations() throws Exception {
...
         enableExtension(TEST_EXTENSION);
 
+        Thread.sleep(10000);
```
pallavi-rao: You could use the waitFor method instead, which uses polling rather than a fixed sleep time.

unit/src/test/java/org/apache/falcon/unit/TestFalconUnit.java
sandeepSamudrala: Added the test case for the same.

pallavi-rao:
Merging this as this is a blocker for 0.11 release. If anyone has any further comments, please raise a JIRA.

PR #386, opened 2017-08-28
FALCON-2304 SchedulerServiceTest.testDeRegistration test fails

pravin-dsilva:
The test fails with

```
java.lang.AssertionError: expected:<1> but was:<null>
	at org.testng.Assert.fail(Assert.java:89)
	at org.testng.Assert.failNotEquals(Assert.java:489)
	at org.testng.Assert.assertEquals(Assert.java:118)
	at org.testng.Assert.assertEquals(Assert.java:160)
	at org.apache.falcon.notification.service.SchedulerServiceTest.testDeRegistration(SchedulerServiceTest.java:274)
```
The test passed by increasing the timeout.

### DIFF
--- a/scheduler/src/test/java/org/apache/falcon/notification/service/SchedulerServiceTest.java
+++ b/scheduler/src/test/java/org/apache/falcon/notification/service/SchedulerServiceTest.java
@@ -269,7 +269,7 @@ public class SchedulerServiceTest extends AbstractTestBase {
         stateStore.putExecutionInstance(new InstanceState(instance3));
         scheduler.unregister(handler, instance3.getId());
 
-        Thread.sleep(100);
+        Thread.sleep(300);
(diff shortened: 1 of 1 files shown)


pallavi-rao (changes requested):
@pravin-dsilva, thanks for picking this up. A better solution for such async methods, would be to have a simple method that does sleep-check loop till a given timeout. Just a sleep is not reliable as it would take different times in different environments.  Something on the lines of FalconUnitTestBase.waitFor.

Having said that, your change does not really impact the test adversely, so I can merge this now and you can implement a proper solution in a separate JIRA. Let me know how you would like to proceed.

Prajyot-Parab:
@pallavi-rao Have raised a new PR - https://github.com/apache/falcon/pull/416 with waitFor implementation to fix the issue.

PR #388, opened 2017-09-04
FALCON-2305 workflow metrics thread failures

PraveenAdlakha:


prism/src/main/java/org/apache/falcon/service/BacklogMetricEmitterService.java
```diff
@@ -417,8 +417,14 @@ public void run() {
...
-                                            new Date(nominalTime.getTime() + 200), PROCESS_LIFE_CYCLE, false);
+                                    InstancesResult status = null;
+                                    try {
+                                        status = wfEngine.getStatus(entity, nominalTime,
+                                                new Date(nominalTime.getTime() + 200), PROCESS_LIFE_CYCLE, false);
+                                    } catch (FalconException e) {
+                                        LOG.error("Unable to get status " + entity.getName() , e);
+                                        continue;
+                                    }
                                     if (status.getInstances().length > 0
```
peeyushb: Are there chances that status remain null. If yes, then we should have a check for status is null or not.

PraveenAdlakha: donot think so from the code path either it will throw exception or will return the status but to be on safer side will add the null check.

pallavi-rao:
Dev test results, please?

pallavi-rao:
@PraveenAdlakha, please also fix the checkstyle warning in the build

pallavi-rao:
@peeyushb, can you please re-review the changes? Need to merge this soon as we are facing this issue in production.

pallavi-rao:
Seeing that the comments have been addressed, merging this as QA is blocked. @peeyushb, if you have any more concerns, it can taken up as a separate JIRA.

PR #391, opened 2017-12-13
FALCON-2317 : Update NOTICE to include references to JMS spec license

gzsombor:
Update NOTICE


gzsombor:
I don't think that this change causing the build failure, can we merge it anyway ?

PR #394, opened 2018-01-05
FALCON-2320 Workaround for "Falcon is not buildable/testable with mav…

dionusos:
…en 3.5.1+"

Change-Id: I0394cffdc9101ebbcc26acb99a3408aaeb317278


pallavi-rao:
Thanks @dionusos. However, @sandeepSamudrala has addressed this as part of PR 393 which is now merged.

PR #395, opened 2018-02-16
FALCON-2321 When Prism is Kerberized, it doesn't pass its token while making calls to Falcon Server

pallavi-rao:


prism/src/main/java/org/apache/falcon/resource/channel/HTTPChannel.java
```diff
@@ -68,6 +80,18 @@
     private String serviceName;
     private Class service;
 
+    /* Name of the HTTP cookie used for the authentication token between the client and the server.
```
sandeepSamudrala: Forwarding the http cookie from the client to server via prism?

PR #396, opened 2018-02-19
FALCON-2322 cast Action.getAny() to org.w3c.dom.Node instead of Eleme…

bpgergo:
…ntNSImpl in order to eliminate ClassCastException


bpgergo:
Closing and recreating this

PR #397, opened 2018-02-20
FALCON-2323 upgrade FasterXML Jackson to 2.9.4

bpgergo:


sandeepSamudrala:
Whats the rational behind updating the jackson version? The build failure is not related to your changes.

bpgergo:
@sandeepSamudrala initially the rational behind the version bumps was to avoid security vulnerabilities.

I added Spring version bump for the same reason.

I also noticed, that the Jackson version bump may eliminate some strange errors which occur during UT runs.

PR #399, opened 2018-02-21
Falcon-2324 Retry, Latererun and JobLog mover services not working with kerberos enabled

sandeepSamudrala:


oozie/src/main/java/org/apache/falcon/workflow/engine/OozieClientFactory.java
```diff
@@ -55,7 +55,7 @@ public static OozieClient get(String clusterName) throws FalconException {
         return get((Cluster) ConfigurationStore.get().get(EntityType.CLUSTER, clusterName));
     }
 
-    private static OozieClient getClientRef(String oozieUrl)
+    public static OozieClient getClientRef(String oozieUrl)
```
pallavi-rao: Instead of exposing this method, JobLogMover should use OozieClientFactory.get(context.getClusterName).

PR #401, opened 2018-02-23
Falcon 2326 Falcon Pull Request Builds failing

pallavi-rao:
Two clashing versions of AbstractFalconAspect were coming in: One from falcon-cli and another from falcon-prism. The fix is to exclude falcon-metrics from falcon-cli dependency.

While I was at it, I also upgraded aspectj libraries, maven plugin and the java version.


pallavi-rao:
See PR 402

PR #406, opened 2018-03-05
FALCON-2330 Update Notice copyright period to 2018

sandeepSamudrala:


pallavi-rao:
Merged without much notice as this is blocking the release and change is trivial

PR #411, opened 2018-05-11
FALCON-2338 EntityGraphTest.initConfigStore fails intermittently

sonia-garudi:
The error occurs only when the MetadataMappingServiceTest testsuite runs before the testcase mentioned above. This test suite creates a file in target folder 'target/store/PROCESS/sample-process.xml' which isn't deleted post the test. 

The change proposed in this pull request deletes the files and the tests in EntityGraphTest suite passed.

@pallavi-rao could you please check this? thanks!


sonia-garudi:
The test failures seen are not related to the changes applied by the patch.

yussufsh:
@pallavi-rao any update on this PR and FALCON-2338 ?

PR #415, opened 2018-07-31
FALCON-2341 Entity SLA Alert publishing wrong results into DB

amitkhanna1806:


prism/src/main/java/org/apache/falcon/service/EntitySLAAlertService.java
```diff
@@ -145,8 +145,9 @@ void processSLACandidates(){
                     if (entityType.equalsIgnoreCase(EntityType.PROCESS.name())){
                         store.putSLAAlertInstance(entityName, clusterName, entityType,
                                 nominalTime, true, false);
```
pallavi-rao: Shouldn't the final arg of putSLA... be true, as the critical watermark is breeched too?

pallavi-rao (changes requested):

prism/src/main/java/org/apache/falcon/service/EntitySLAAlertService.java
amitkhanna1806: Agreed.. Made The change..

pallavi-rao (approved):
Thanks @amitkhanna1806

PR #416, opened 2018-09-19
FALCON-2342 testDeRegistration timeout fixed using waitFor function

Prajyot-Parab:
The test fails with

```
java.lang.AssertionError: expected:<1> but was:<null>
	at org.testng.Assert.fail(Assert.java:89)
	at org.testng.Assert.failNotEquals(Assert.java:489)
	at org.testng.Assert.assertEquals(Assert.java:118)
	at org.testng.Assert.assertEquals(Assert.java:160)
	at org.apache.falcon.notification.service.SchedulerServiceTest.testDeRegistration(SchedulerServiceTest.java:274)
```

Implemented waitFor() function to fix it.


Prajyot-Parab:
@pallavi-rao Have raised this new PR -  with waitFor implementation to fix the issue discussed in #386. Can you please take a look at this.

